PR #22, opened 2018-03-19
Update license check to check for specific copyright year.

pasinskim:
License test is checking now when the source file was modified and
makes sure the last modification date matches the copyright information
inside the source file.
This way all the modified files need to have the license updated, but
the ones which are not modified can stay with the previous copyright
information.

Changelog: None

Signed-off-by: Marcin Pasinski <marcin.pasinski@northern.tech>

### DIFF
--- a/check_license_go_code.sh
+++ b/check_license_go_code.sh
@@ -1,8 +1,6 @@
 #!/bin/bash
 
 cat > license.tmp <<EOF
-// Copyright 2017 Northern.tech AS
-//
 //    Licensed under the Apache License, Version 2.0 (the "License");
 //    you may not use this file except in compliance with the License.
 //    You may obtain a copy of the License at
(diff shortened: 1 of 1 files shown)


check_license_go_code.sh
```diff
@@ -16,16 +14,29 @@ cat > license.tmp <<EOF
...
+# we need to add two extra lines missing from the license preamble
+# // Copyright <copyrigt_year> Northern.tech AS
+# //
+lines=$(($lines + 2))
 
 ret=0
 for each in $(find . -type f \( ! -regex '.*/\..*' ! -path "./Godeps/*" ! -path "./vendor/*" -name '*.go' \)); do
-  echo "Checking $each for correct license header"
-  head -n $lines $each | diff -qu license.tmp - > /dev/null
+  modified_year=$(date +%Y -r $each)
```
kacf: You cannot check it like this, because git always checks out all files at today's date. Try this instead:

```
git log -n1 --format=%ad --date=format:%Y -- $each
```

check_license_go_code.sh
```diff
@@ -16,16 +14,29 @@ cat > license.tmp <<EOF
...
     echo "!!! FAILED license check on $each"
     ret=1
   else
-    echo "License check passed on $each"
+    copyright_modified=$(echo "// Copyright <copyright_year> Northern.tech AS" | sed "s/<copyright_year>/$modified_year/g")
+    copyright_file=$(head -n 1 $each)
+    if [ "$copyright_modified" == "$copyright_file" ]; then
+      echo "License check passed on $each"
+    else
+      echo "!!! FAILED license check on $each; copytight year does not match modification year"
```
kacf: Propose switching `copytight year does not match modification year` with `make sure copyright year matches last modified year of the file`.

check_license_go_code.sh
pasinskim: Thanks for spotting this. Fixed!

check_license_go_code.sh
pasinskim: It's your baby, so will do as you want :-)

pasinskim:
@kacf all fixed now. Can you recheck?

kacf:
Perfect, thanks!

PR #26, opened 2019-02-19
Update license year and fix Travis build

lluiscampos:
The main change is the logic to deduce the "commit range" on where to run the commit checks.

See individual commits for more info on the other required changes.


check_commits.sh
```diff
@@ -9,9 +9,16 @@ case "$1" in
...
+if [ -n "$TRAVIS_PULL_REQUEST_BRANCH" -a -n "$TRAVIS_BRANCH" ]
 then
+    # It is an actual pull request
     COMMIT_RANGE="$TRAVIS_BRANCH..HEAD"
+elif [ -n "$TRAVIS_COMMIT_RANGE" ]
+then
+    # It is a regular branch build
+    COMMIT_RANGE="$TRAVIS_COMMIT_RANGE"
+    # Workaround for force-pushed branches
+    git rev-list --no-merges "$COMMIT_RANGE" 2>/dev/null >/dev/null || COMMIT_RANGE=HEAD~1..HEAD
```
kacf: I don't get it. Why is this workaround needed?

lluiscampos: TRAVIS_COMMIT_RANGE will be a range from the last build to the current one. So this is a workaround for when the previous tested commit is not there anymore. This happens when you force push modifying old commits.

I cannot come with a good example, but it happened to me during development that I amended a commit and then in the next build travis was using it in the TRAVIS_COMMIT_RANGE and the script was failing.

kacf: But then why use `TRAVIS_COMMIT_RANGE` at all? Is there a case where `TRAVIS_BRANCH` is not set?

FWIW, testing for `TRAVIS_PULL_REQUEST_BRANCH`, and hence running the test only on pull request branches, is intentional. The thing is that once something has been pushed to master, it is no longer fixable, hence we don't want this check to run for any commits on a non-PR travis branch.

lluiscampos: > But then why use `TRAVIS_COMMIT_RANGE` at all? Is there a case where `TRAVIS_BRANCH` is not set?

`TRAVIS_BRANCH` is always set.

> FWIW, testing for `TRAVIS_PULL_REQUEST_BRANCH`, and hence running the test only on pull request branches, is intentional. The thing is that once something has been pushed to master, it is no longer fixable, hence we don't want this check to run for any commits on a non-PR travis branch.

Maybe this PR is not adding much value after all if we reject this change.

My reasoning was that while working on integrating the commit checks in mender-artifact I wanted first my **branch**  to fail with a commit that had no Changelog. But then I realized that the script used range `my-branch..HEAD` which is 0 commits. So I thought introducint TRAVIS_COMMIT_RANGE for this cases.

I don't have an strong opinion. I can remove the change and we merge only the other minor things

kacf: Yes, I see your point, but I think we have to drop this because it can break master builds and release builds, if someone by mistake pushes a commit that doesn't pass.

lluiscampos: Ok. Removing this commit.

PR #29, opened 2019-07-11
Fix commit range detection for Gitlab builds yet again.

kacf:
Changelog: None

Signed-off-by: Kristian Amlie <kristian.amlie@northern.tech>


lluiscampos:
Nice work ;)

PR #30, opened 2019-10-07
Set checksum filepath to default value

alfrunes:
Made it possible to modify the CHKSUM_FILE variable if it's defined in
the environment.

Signed-off-by: Alf-Rune Siqveland <alf.rune@northern.tech>

### DIFF
--- a/check_license.sh
+++ b/check_license.sh
@@ -21,7 +21,7 @@ fi
 # Check license of dependencies.
 ################################################################################
 
-CHKSUM_FILE=LIC_FILES_CHKSUM.sha256
+CHKSUM_FILE=${CHKSUM_FILE:-LIC_FILES_CHKSUM.sha256}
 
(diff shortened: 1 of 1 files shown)


kacf:
What's the motivation for this change? This looks a bit "debug-gy" to me.

alfrunes:
I moved the `LIC_FILES_CHKSUMS.sha256` away from the root directory in `mender` [#439](https://github.com/mendersoftware/mender/pull/439) into the `vendor` directory.

kacf:
Why? I would prefer not to change this as it is standard in all our Go, and Yocto expects it in this location as well.

alfrunes:
You're right. I was just trying to move as many "unneeded" files from the root directory in the `mender` repository for esthetic reasons :). However, if Yocto depends on this file location I'll revert these changes and close this PR.

PR #31, opened 2019-10-25
Create Pipeline(s) for mendertesting  repo

lluiscampos:
These independent check-commits and check-license pipelines are meant to
be included by any other GitLab Pipeline in the organization.

Note that from Travis to GitLab, the "go build" step has been removed.
This repo does not produce any binary so the build is unnecesary.

Changelog: None

### DIFF
--- a/.gitlab-ci-check-license.yml
+++ b/.gitlab-ci-check-license.yml
@@ -1,0 +1,23 @@
+
+stages:
+  - test
+
+test:check-license:
+  image: alpine
+  stage: test
+  before_script:
+    # Install dependencies
+    - apk add --no-cache git bash perl-utils
+    # Rename the branch we're on, so that it's not in the way for the
+    # subsequent fetch. It's ok if this fails, it just means we're not on any
+    # branch.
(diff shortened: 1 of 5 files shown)


.gitlab-ci-check-license.yml
```diff
@@ -0,0 +1,23 @@
...
+    - git branch -m temp-branch || true
+    # Git trick: Fetch directly into our local branches instead of remote
+    # branches.
+    - git fetch origin 'refs/heads/*:refs/heads/*'
+    # Get last remaining tags, if any.
+    - git fetch --tags origin
+    - git clone http://github.com/mendersoftware/mendertesting /tmp/mendertesting
+  script:
+    # Check licenses
+    - /tmp/mendertesting/check_license.sh
```
kacf: Both this test and the `check_commits.sh` are duplicated, since they are already running in the unit test stage. Is the intention here that this be reusable downstream?

kacf: Bah, never mind, you already answered it. Sorry, I wasn't part of the original review. Moving on...

PR #33, opened 2019-11-08
QA-111: Check signed commits on community repositories 

oleorhagen:
This is done through adding two flags to the chech_commits.sh script.
One for enabling the signature check, and one for enabling the Changelog check.

In the case that no flags are given, they will both be checked (The old
behaviour).

Also add the ability to pass these parameters to the
'.gitlab-ci-check-commits.yml' file through an environment variable
'MENDERTESTING_ARGS'.

Changelog: None

### DIFF
--- a/.gitlab-ci-check-commits-changelogs.yml
+++ b/.gitlab-ci-check-commits-changelogs.yml
@@ -1,0 +1,23 @@
+
+stages:
+  - test
+
+test:check-commits:
+  image: alpine
+  stage: test
+  before_script:
+    # Install dependencies
+    - apk add --no-cache git bash
+    # Rename the branch we're on, so that it's not in the way for the
+    # subsequent fetch. It's ok if this fails, it just means we're not on any
+    # branch.
+    - git branch -m temp-branch || true
+    # Git trick: Fetch directly into our local branches instead of remote
+    # branches.
+    - git fetch origin 'refs/heads/*:refs/heads/*'
+    # Get last remaining tags, if any.
+    - git fetch --tags origin
+    - git clone http://github.com/mendersoftware/mendertesting /tmp/mendertesting
+  script:
+    # Verify that the commits have a changelog.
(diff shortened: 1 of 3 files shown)


oleorhagen:
@lluiscampos Initial script with an opportunity to run only changelog or signature checks. Also there is the ability to pass the flags on to the script in 'gitlab-ci-check-commits.yml'. 

I'm a little unsure as to how this integrates with the tests in the local repos though. I see you've commented that the stretch-goal is done. Does that mean that this is all it takes, and the flags has to be passed from 'mender-qa' on a per-repository basis? 

Also, does the task include removing the vendored mendertesting repository from all the mendersoftware repositories?

As far as I was able to track down, it is these ones:

deployments
deployments-enterprise
deviceadm
deviceauth
go-lib-micro
inventory
log
mender-artifact
mender-cli
mender-stress-test-client
useradm
useradm-enterprise

.gitlab-ci-check-commits.yml
```diff
@@ -20,4 +20,4 @@ test:check-commits:
     - git clone http://github.com/mendersoftware/mendertesting /tmp/mendertesting
   script:
     # Check commit compliance.
-    - /tmp/mendertesting/check_commits.sh
+    - /tmp/mendertesting/check_commits.sh ${MENDERTESTING_ARGS}
```
lluiscampos: How is this variable supposed to be used? Will that actually work across pipelines? I think the best would be to create two new templates: `.gitlab-ci-check-commits-signoffs.yml` and `...-changelogs.yml`

lluiscampos:
> Also, does the task include removing the vendored mendertesting repository from all the mendersoftware repositories?

No. But I see now that your question comes from a comment in the JIRA ticket. That was something pointed out in a QA meeting but later on discarded for the above reason: that we must have them vendored to be able to correctly check on the enterprise forks.

Sorry that this was not clear on the JIRA side

oleorhagen:
@lluiscampos should be good for another look now

PR #35, opened 2019-12-20
Make the check_python_code script not error when no files are wrong

oleorhagen:
Quote the strings in the zero-length string test.

Changelog: None


oleorhagen:
To be honest, I'm not entirely sure why this is needed. Not how I thought '-n' was working.

PR #37, opened 2019-12-27
Skip master execution of mendertesting commit checks

lluiscampos:
This is mainly for mender-qa, where there vast majority of times we are
using "master" and wasting precious GitLab CI minutes with commit
checks. It won't hurt saving minutes on other repos neither.

Strictly speaking, we could apply the same principle to all templates,
but let's limit it only to the commit ones: as these are the really
superfluous ones in master while the others could possibly detect
errors in ninja merges.

### DIFF
--- a/.gitlab-ci-check-commits.yml
+++ b/.gitlab-ci-check-commits.yml
@@ -3,8 +3,10 @@ stages:
   - test
 
 test:check-commits:
-  image: alpine
   stage: test
+  except:
+    - /^(master|[0-9]+\.[0-9]+\.x)$/
+  image: alpine
   before_script:
     # Install dependencies
     - apk add --no-cache git bash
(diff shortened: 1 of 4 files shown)


lluiscampos:
In a second thought, I reworked this to only limit the commit checks (in case you have seen the original PR).

.gitlab-ci-check-commits.yml
```diff
@@ -3,8 +3,10 @@ stages:
   - test
 
 test:check-commits:
-  image: alpine
   stage: test
+  except:
+    - master
```
kacf: Should these actually have the famous `/^(master|[0-9]+\.[0-9]+\.x)$/` expression? Not sure if we run nightly on those, but definitely they have the same limitation as the master branch: Commit checks are useless on them.

.gitlab-ci-check-python3-format.yml
```diff
@@ -5,7 +5,7 @@
 # not formatted according to standard Black formatting rules.
 #
 # Add it to the project in hand through Gitlab's include functionality
-# 
+#
```
oleorhagen: :+1:

.gitlab-ci-check-commits.yml
lluiscampos: Yes! It is not critical but rather a good practice. I'll fix.

PR #43, opened 2020-02-23
QA-149: Use pipelines templates in all back-end repos

lluiscampos:
This PR collects a set of changes and extensions for the templates to work with all repositories. It also has as important bugfix!

Best review commit by commit, I think.

Important: two templates have been renamed. Fortunately, only one repo is using them (inventory) so I will coordinate the merges to not affect ongoing PRs.

This PR will then unblock the following PRs:
* https://github.com/mendersoftware/deviceauth/pull/310
* https://github.com/mendersoftware/tenantadm/pull/121
* https://github.com/mendersoftware/inventory/pull/153
* https://github.com/mendersoftware/deployments/pull/393
* https://github.com/mendersoftware/useradm/pull/151

### DIFF
--- a/.gitlab-ci-check-docker-build.yml
+++ b/.gitlab-ci-check-docker-build.yml
@@ -1,4 +1,4 @@
-# .gitlab-ci-check-golang-dockerimage.yml
+# .gitlab-ci-check-docker-build.yml
 #
 # This gitlab-ci template builds a Docker image assuming a single Dockerfile
 # and publish it on Docker Hub
@@ -9,7 +9,7 @@
 #
 # include:
 #   - project: 'Northern.tech/Mender/mendertesting'
-#     file: '.gitlab-ci-check-golang-dockerimage.yml'
+#     file: '.gitlab-ci-check-docker-build.yml'
 #
 # Requires the following variables set in the project CI/CD settings:
 # DOCKER_HUB_USERNAME: Username for docker.io
--- a/.gitlab-ci-check-docker-build.yml
+++ b/.gitlab-ci-check-docker-build.yml
@@ -36,8 +36,8 @@ build:
     - docker:19.03.5-dind
   before_script:
     - export DOCKER_TAG=${CI_COMMIT_REF_SLUG:-master}
-    - export SERVICE_IMAGE="$DOCKER_REPOSITORY:$DOCKER_TAG"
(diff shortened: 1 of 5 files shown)


lluiscampos:
@mzedel This PR is ready for review. If you have some time during your trip, I would very much appreciate it! All the related PRs are also ready, but if you don't have much time focus on this one only.

Thanks.

.gitlab-ci-check-docker-build.yml
```diff
@@ -28,10 +35,9 @@ build:
   services:
     - docker:19.03.5-dind
   before_script:
-    - export DOCKER_REPOSITORY="mendersoftware/${CI_PROJECT_NAME}"
     - export DOCKER_TAG=${CI_COMMIT_REF_SLUG:-master}
     - export SERVICE_IMAGE="$DOCKER_REPOSITORY:$DOCKER_TAG"
```
mzedel: I'm not sure if this would ever not apply, or if it is just searching for alignments... but might `"$DOCKER_REPOSITORY":"$DOCKER_TAG"` be necessary here as well - if the `COMMIT_TAG` is built similarly?

lluiscampos: It is not necessary as `: ` is not a valid character for bash variables while `_` is. However, you already know how much I love alignment, so I will quote these ones here also.

lluiscampos:
>     * would it be helpful to extract the enterprise acceptance tests, to their own jobs at some point (based on the presence of the env vars or the compose file)?

This might make sense as it will be explicit when an Enterprise test fails.

However, I've been thinking a bit about the consequences wrt code coverage reporting... And then I realized that we might actually be losing the code coverage fro the first os run as it is now! I created bug [QA -159](https://tracker.mender.io/browse/QA-159) to follow up.

>     * would there be any way around the liblzma dependency? (just since the apt steps always take a bit)

Yes, it is a shame that it is only used in `deployments` but installed for all.

I have an idea; each repo to could have an OS dependencies files (similar to [these](https://github.com/mendersoftware/integration/tree/master/tests/requirements)) so that deployments could specify liblzma and the rest none. The template basically does if such file exists, then install stuff. What do you think?

lluiscampos:
> I have an idea; each repo to could have an OS dependencies files (similar to [these](https://github.com/mendersoftware/integration/tree/master/tests/requirements)) so that deployments could specify liblzma and the rest none. The template basically does if such file exists, then install stuff. What do you think?

I reworked the templates to do this.

PR #48, opened 2020-03-19
Pass git commit hash to the docker build

merlin-northern:
Proposition to pass the git commit to the build

* goes with https://github.com/mendersoftware/useradm-enterprise/pull/43

ChangeLog:title


merlin-northern:
> I would be nice to document this among the rest of the template documentation. Maybe in the future we add more variables passed to the Dockerfile(s), so it would be useful for the user to get them listed in the doc together with the rest of the information.
> 
> LGTM 👍

thanks!
follow up: https://github.com/mendersoftware/mendertesting/pull/49

PR #54, opened 2020-04-02
Templates Docker publish related fixes

lluiscampos:
- [templates] Take password from CLI on docker login calls
The --password-stdin option is throwing errors in some GitLab shared
runners. The password will be masked anyway, so it is safe.

- [docker-build-template] Log in in both registries inconditionally
Now all repos have both credentials anyway (inheriged from the group) so
just log in in both.

- [docker-build-template] Publish Enterprise images to Docker Hub
Add an extra job for Enterprise repo (i.e. when the caller has
registry.mender.io in DOCKER_REPOSITORY) to publish the image with the
same tags into Docker Hub.

### DIFF
--- a/.gitlab-ci-check-docker-build.yml
+++ b/.gitlab-ci-check-docker-build.yml
@@ -73,10 +73,7 @@ publish:image:
     - docker load -i image.tar
     - docker tag $SERVICE_IMAGE $DOCKER_REPOSITORY:$COMMIT_TAG
     - docker tag $SERVICE_IMAGE $DOCKER_REPOSITORY:$CI_COMMIT_REF_SLUG
-    - if [ -n "$REGISTRY_MENDER_IO_USERNAME" ]; then
-        docker login -u $REGISTRY_MENDER_IO_USERNAME -p $REGISTRY_MENDER_IO_PASSWORD registry.mender.io;
-      else
-        docker login -u $DOCKER_HUB_USERNAME -p $DOCKER_HUB_PASSWORD;
-      fi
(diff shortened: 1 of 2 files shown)


lluiscampos:
@mzedel Please pay close attention to the last commit. I have not tested it.

lluiscampos:
> @mzedel Please pay close attention to the last commit. I have not tested it.

I did CI lint it, though. I know for sure that inheritance between the template and the jobs work as expected (with regards to the only refs + only variables). What I need your close eye is on the variables manipulation for `docker tag` + `docker push`

.gitlab-ci-check-docker-build.yml
```diff
@@ -76,4 +78,22 @@ publish:image:
...
+  extends: .publish_image
+
+# Extra job for Enterprise repos, publish also to Docker Hub
+publish:image:dockerhub:
+  extends: .publish_image
+  only:
+    variables:
+      - $DOCKER_REPOSITORY =~ /^registry\.mender\.io.\/*/
+  before_script:
+    # The trick here is to let SERVICE_NAME use the origianl DOCKER_REPOSITORY (i.e. registry.mender.io)
```
mzedel, line 94: the only question I have is how the `SERVICE_NAME` image is now pushed, since both  calls refer to the `DOCKER_REPOSITORY` now?

lluiscampos: `SERVICE_NAME` is not pushed, only the "other" tags (which in fact will be the same name). I think it will work, but I guess @tranchitella will report if I am wrong :)

PR #60, opened 2020-04-20
[github status] Do not run for master and other protected branches

lluiscampos:
- [github status] Do not run for master and other protected branches
The motivation for this change is to skip them for mender-qa master,
which would report to non existing PR (master) when running integration
pipeline in behalf of a PR in any other repo.

- [commit checks] Add staging to the list of branches to skip
Just to keep it aligned with our current "protected branches".

### DIFF
--- a/.gitlab-ci-github-status-updates.yml
+++ b/.gitlab-ci-github-status-updates.yml
@@ -13,6 +13,8 @@ stages:
 
 github:start:
   stage: .pre
+  except:
+    - /^(master|staging|[0-9]+\.[0-9]+\.x)$/
   # Keep overhead low by using a small image with curl preinstalled.
   image: appropriate/curl
   script:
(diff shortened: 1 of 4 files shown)


.gitlab-ci-github-status-updates.yml
```diff
@@ -13,6 +13,8 @@ stages:
 
 github:start:
   stage: .pre
+  except:
+    - /^(master|staging|[0-9]+\.[0-9]+\.x)$/
```
kacf, line 17: I think we should keep it running here. This is occasionally useful information. For example, after the nightly has run, the status will correctly show up on that commit.

lluiscampos: Okey, that is a valid point. I will then override these only for mender-qa pipeline (expect a new PR later)

PR #61, opened 2020-05-04
QA-161

oleorhagen:
Two parts:

* Fail the test in case of errors
* Source the `check_<functinoality>` scripts, instead of directly executing them.

I think they should be pretty self-explanatory.

### DIFF
--- a/licenses.go
+++ b/licenses.go
@@ -14,13 +14,17 @@
 
 package mendertesting
 
-import "errors"
-import "os/exec"
-import "fmt"
-import "os"
-import "path"
-import "strings"
-import "testing"
+import (
+	"errors"
+	"fmt"
+	"os"
+	"os/exec"
+	"path"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
 
 const packageLocation string = "github.com/mendersoftware/mendertesting"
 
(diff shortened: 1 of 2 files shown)


licenses.go
```diff
@@ -41,43 +41,42 @@ func SetFirstEnterpriseCommit(sha string) {
 	firstEnterpriseCommit = sha
 }
 
-func CheckMenderCompliance(t *testing.T) error {
+func CheckMenderCompliance(t *testing.T) {
```
lluiscampos, line 44: Good! But you need to fix [the tests](https://github.com/mendersoftware/mendertesting/blob/master/license_test.go#L53).

licenses.go
```diff
@@ -41,43 +41,42 @@ func SetFirstEnterpriseCommit(sha string) {
...
 	pathToTool, err := locatePackage()
 	if err != nil {
-		return err
+		t.Errorf("Failed to locate package: %s", err)
 	}
 
-	cmdString := []string{path.Join(pathToTool, "check_license_go_code.sh")}
+	args := []string{"-c", fmt.Sprintf("source %s", path.Join(pathToTool, "check_license_go_code.sh"))}
```
lluiscampos, line 50: I don't understand this. Why do you want to source the script? By looking at the script it will work, but I think it is a bad practice: `source` should be used upon files containing functions or env variables, not actual bash code.

I though that this would be as simple as calling `bash [script]` (i.e. `bash check_license_go_code.sh`), wouldn't that work?

oleorhagen: Ur right, that is way easier! Thx

:racehorse:

oleorhagen:
@lluiscampos Fixed up and squashed into one commit. The codecov I don't really understand. From looking at it it is the Enterprise code check which changes it, but this is never tested for the error case from before.

kacf:
I think the codecov is correct, it's just more lines now. Since it's preexisting I wouldn't sweat it too much. This is a small repository, so the percentage is unfairly big!

PR #65, opened 2020-05-05
[pipeline] Add GitLab CI linting for yml files here

lluiscampos:
So that we prevent breaking the repos using these templates.


lluiscampos:
See API call here: https://docs.gitlab.com/ee/api/lint.html

PR #68, opened 2020-05-14
[docker-build-template] Skip publish:image:mender for non release repos

lluiscampos:
For a repo using the template which is not part of any Mender release,
skip the actual tagging/publishing of mender-VERSION kind of tag.

Changelog: None


lluiscampos:
@kacf This should fix the CI failure we are having in `sre-tools` repo

PR #69, opened 2020-05-15
Fix error assertion when check_license.sh fails

merlin-northern:
When one of the licenses did not match, there was an error:

`panic: interface conversion: interface {} is *mendertesting.MenderComplianceError, not string`

ChangeLog:none

### DIFF
--- a/licenses.go
+++ b/licenses.go
@@ -48,7 +48,11 @@ func SetFirstEnterpriseCommit(sha string) {
 func CheckMenderCompliance(t *testing.T) {
 	t.Run("Checking Mender compliance", func(t *testing.T) {
 		err := checkMenderCompliance()
-		assert.NoError(t, err, err)
+		msg := ""
+		if err != nil {
+			msg = err.Error()
+		}
+		assert.NoError(t, err, msg)
 	})
 }
 


licenses.go
```diff
@@ -48,7 +48,11 @@ func SetFirstEnterpriseCommit(sha string) {
 func CheckMenderCompliance(t *testing.T) {
 	t.Run("Checking Mender compliance", func(t *testing.T) {
 		err := checkMenderCompliance()
-		assert.NoError(t, err, err)
```
alfrunes, line 51: I don't see how this should fix this. assert.NoError is more or less the same as assert.Nil, but there might be a bug in the particular version of testifty/assert for this repo. Have you tried updating the vendored version?

kacf: I think the last `err` should just be removed. I think it will tell you what the error is by using the first argument.

oleorhagen: This fails on older versions of testify. So upgrading testify in the given repo should fix it.

oleorhagen: And @kacf is right. It is sort of superfluous. The error will show up in the `Error` field anywho.

oleorhagen: I would close this in favour of: https://github.com/mendersoftware/mendertesting/pull/71

oleorhagen:
Closed in favour of: https://github.com/mendersoftware/mendertesting/pull/71

licenses.go
merlin-northern: it does fix this. and passing err as last arg was a mistake in the current testify version. Ole said taht it is better to upgrade testify, which fixes this as well. the third arg is a message and passing message is the right thing to do, but skipping it is also all right, as you say.

merlin-northern: the problem with removing the last arg is: you probably will never know which license failed.

merlin-northern: > I would close this in favour of: #71

no problema.

PR #70, opened 2020-05-15
[docker-build-template] Skip publish:image:mender for not found repos

lluiscampos:
Yet another case where we need to skip the publish is when the repo is
not found at all.

### DIFF
--- a/.gitlab-ci-check-docker-build.yml
+++ b/.gitlab-ci-check-docker-build.yml
@@ -122,10 +122,15 @@ publish:image:mender:
     # Get release_tool:
     - git clone https://github.com/mendersoftware/integration.git mender-integration
     - alias release_tool=$(realpath mender-integration/extra/release_tool.py)
-    # If the repo is not part of a Mender release, ignore
+    # If the repo is not recognized, ignore
+    - if ! echo $(release_tool --list git --all) | grep $CI_PROJECT_NAME; then
+    -  echo "Repository $CI_PROJECT_NAME not found in release_tool. Exiting"
+    -  exit 0
+    - fi
+    # If the repo/branch is not part of a Mender release, also ignore
     - integration_versions=$(release_tool --integration-versions-including $CI_PROJECT_NAME --version $CI_COMMIT_REF_SLUG | sed -e 's/origin\///')
     - if test -z "$integration_versions"; then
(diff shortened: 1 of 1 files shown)


lluiscampos:
The previous PR #68 was not enough to fix `sre-tools` build failure.

.gitlab-ci-check-docker-build.yml
```diff
@@ -122,10 +122,15 @@ publish:image:mender:
...
+    - if ! echo $(release_tool --list git --all) | grep $CI_PROJECT_NAME; then
+    -  echo "Repository $CI_PROJECT_NAME not found in release_tool. Exiting"
+    -  exit 0
+    - fi
+    # If the repo/branch is not part of a Mender release, also ignore
     - integration_versions=$(release_tool --integration-versions-including $CI_PROJECT_NAME --version $CI_COMMIT_REF_SLUG | sed -e 's/origin\///')
     - if test -z "$integration_versions"; then
-    -  echo "Repository $CI_PROJECT_NAME version $CI_COMMIT_REF_SLUG is not part of any Mender release"
+    -  echo "Repository $CI_PROJECT_NAME version $CI_COMMIT_REF_SLUG is not part of any Mender release. Exiting"
     -  exit 0
```
kacf, line 134: I'm confused, why isn't the `test -z "$integration_versions"` check enough? Shouldn't that be empty if the repository is not recognized?

The `--list` call only checks the master branch of `integration`, so it will miss if the repository is being used for 2.3, but not for master for example. I'm not sure if this has any practical effect for us, but maybe it will prevent pushing for repositories which are obsoleted, but still in use by old versions.

lluiscampos: Good question. Let's see:

The check is not enough because 1) returns 1 if the component does not exist and (most importantly 2) prints `Unrecognized repository: sre-tools` to stdout. We could grep for this "Unrecognized repository" string and ignore exit code, but I thought that was misusing the tool.

Then, `--list` in combination with `--all` list all components, including the ones marked with `release component: False`, sot that should catch the case when a component is non releasable for master but it was for an old release. For example in master right now lists `mender-conductor` or `deviceadm`, so it should be okey.

kacf: Ah, you're right, the `--all` will cover the case when the repository has only been part of mender *at some point*. Then all is good!

PR #71, opened 2020-05-15
fix: Remove the err message from the NoErr assertion 

oleorhagen:
The error is already printed as the Error message. Also the err as the message
panics on older versions of testify.

Changelog: None


lluiscampos:
Supersedes #69?

PR #72, opened 2020-05-19
[docker-build-template] REF_NAME instead of REF_SLUG for publish tags

lluiscampos:
So that branch 1.2.x of a given repo gets published as <repo>:1.2.x (and
mender-1.2.x) instead of 1-2-x.

Renamed a bit the env variables to clarify their usage.

Obs! This will not work for publishing branches with slash (like
feature/my-coool-stuff) but that we can fix later on.

### DIFF
--- a/.gitlab-ci-check-docker-build.yml
+++ b/.gitlab-ci-check-docker-build.yml
@@ -27,13 +27,18 @@
 #
 # Passes the following variables to the build:
 #   * GIT_COMMIT_TAG
-#     * SHA of git commit at which build occured
+#     * SHA of git commit at which build ocurred
 #     * passed via: --build-arg GIT_COMMIT_TAG="${COMMIT_TAG}"
 #
 
 stages:
   - build
   - publish
+
+.export_docker_vars: &export_docker_vars |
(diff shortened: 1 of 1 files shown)


.gitlab-ci-check-docker-build.yml
```diff
@@ -27,14 +27,20 @@
...
-#     * passed via: --build-arg GIT_COMMIT_TAG="${COMMIT_TAG}"
+#     * SHA of git commit at which build ocurred
+#     * passed via: --build-arg GIT_COMMIT_TAG="${DOCKER_PUBLISH_COMMIT_TAG}"
 #
 
 stages:
   - build
   - publish
 
+.export_docker_vars: &export_docker_vars |
```
mzedel, line 38: 😍 - I guess here it is "two or more: use a template"

lluiscampos:
> looks reasonable ok_hand
> besides slashes we might have to handle the length as well since I'm not sure if there is a limitation with the docker tags, but we don't have a lot of very long branch names either, so that can also wait for a while slightly_smiling_face

Correct. As per today we only publish master, staging or release branches. We have an specific task to publish feature branches and then we need to fix this somehow.

PR #73, opened 2020-05-25
[docker-build-template] QA-179: Trigger saas sync on staging changes

lluiscampos:
New job to run on <repo>/staging branch. It will trigger saas sync job
once per container in in the repo.

### DIFF
--- a/.gitlab-ci-check-docker-build.yml
+++ b/.gitlab-ci-check-docker-build.yml
@@ -41,6 +41,12 @@ stages:
   export DOCKER_PUBLISH_TAG=${CI_COMMIT_REF_NAME}
   export DOCKER_PUBLISH_COMMIT_TAG=${CI_COMMIT_REF_NAME}_${CI_COMMIT_SHA}
 
+.get_release_tool_alpine: &get_release_tool_alpine |
+  apk add git python3
+  pip3 install pyyaml
+  git clone https://github.com/mendersoftware/integration.git mender-integration
+  alias release_tool=$(realpath mender-integration/extra/release_tool.py)
+
 build:docker:
   stage: build
   tags:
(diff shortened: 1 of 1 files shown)


lluiscampos:
See related PR https://github.com/mendersoftware/saas/pull/327

.gitlab-ci-check-docker-build.yml
```diff
@@ -155,3 +158,31 @@ publish:image:mender:dockerhub:
...
+    - *get_release_tool_alpine
+    - apk add --no-cache curl
+  script:
+    # If the repo is not recognized, ignore
+    - if ! echo $(release_tool --list git --all) | grep $CI_PROJECT_NAME; then
+    -  echo "Repository $CI_PROJECT_NAME not found in release_tool. Exiting"
+    -  exit 0
+    - fi
+    # Trigger the sync once per container in this repo
+    - for container in $(release_tool -m git ${CI_PROJECT_NAME} container); do
```
mchalski, line 179: nice - didn't occur to me to use release-tool in this manner, noted :+1:

.gitlab-ci-check-docker-build.yml
```diff
@@ -155,3 +158,31 @@ publish:image:mender:dockerhub:
...
+    -   echo "Triggering saas staging sync for container $container image $DOCKER_PUBLISH_COMMIT_TAG"
+    -   curl -v -f -X POST
+          -F token=${SAAS_TRIGGER_TOKEN}
+          -F ref=master
+          -F variables[TRIGGER_SYNC_STAGING_COMPONENT]=true
+          -F variables[SYNC_CONTAINER_NAME]=${container}
+          -F variables[SYNC_IMAGE_NAME]=${DOCKER_PUBLISH_COMMIT_TAG}
```
lluiscampos, line 186: Note to self: this is wrong; it should prepend the image name (i.e. `mendersoftware/repo:`)

lluiscampos: Actually not; we cannot add pass the image name here as it will be different for different containers (`workflows` and `workflows-worker`).

So we need to send only the version part and modify the logic in https://github.com/mendersoftware/saas/pull/327.

I'll fix in the next push.

lluiscampos:
Updated the regex here following last rework in https://github.com/mendersoftware/saas/pull/327. The caller will pass only the Docker tag and the logic here shall respect the whatever image name we are using.

As a side effect, this also will take care of a related problem: the enterprise repos are publishing the images to registry.mender.io while saas repo is using the ones from Docker hub. Hence not passing the images will allow saas repo to switch to any other registry at any time.

lluiscampos:
Las push just removed the extra `trigger` stage and used `.post` instead.

This PR is now ready.

PR #77, opened 2020-06-02
[python-formatting-pipeline] skip unnecessary python installation

mzedel:
Changelog: None
Signed-off-by: Manuel Zedel <manuel.zedel@northern.tech>

### DIFF
--- a/.gitlab-ci-check-python3-format.yml
+++ b/.gitlab-ci-check-python3-format.yml
@@ -49,14 +49,13 @@ stages:
   - test
 
 test:check-python3-formatting:
-  image: debian
+  image: python:3.8-alpine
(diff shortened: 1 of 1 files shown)


.gitlab-ci-check-python3-format.yml
```diff
@@ -49,14 +49,12 @@ stages:
   - test
 
 test:check-python3-formatting:
-  image: debian
+  image: python:3.8.3
```
lluiscampos, line 52: Should we just set it to `python:3.8` so that we get bugfixes updates?

mzedel: good point! should now be `3.8` 👌

PR #78, opened 2020-06-03
[check_commits] Allow commits from dependabot to not include changelog

lluiscampos:


check_commits.sh
```diff
@@ -90,15 +90,15 @@ do
...
+    if [ -n ${CHECK_SIGNOFFS} ]; then
         # Check that Signed-off-by tags are present.
         if ! echo "$COMMIT_MSG" | grep -F "Signed-off-by: ${COMMIT_USER_EMAIL}" >/dev/null; then
             echo >&2 "Commit ${i} is not signed off! Use --signoff with your commit."
             notvalid="$notvalid $i"
         fi
     fi
 
-    if [ ! -z $CHECK_CHANGELOGS ]; then
+    if [ -n ${CHECK_CHANGELOGS} -a ${COMMIT_USER_EMAIL} != "support@dependabot.com" ]; then
```
lluiscampos, line 101: This is not the most robust solution; a regular user could set such an email in his/her Git configuration and get commits without changelog. I just don't think that the potential harm of that is worth the time in making this too fancy...

oleorhagen: I think that is gripping for straws. It still has to go through us also.

check_commits.sh
```diff
@@ -90,15 +90,15 @@ do
         continue
     fi
 
-    if [ ! -z $CHECK_SIGNOFFS ]; then
+    if [ -n ${CHECK_SIGNOFFS} ]; then
```
oleorhagen, line 93: Nitpick :straight_ruler: : 
Dirtying blame log

lluiscampos: Hehe :policeman: 
I can argue that this is related: I did modify `! -z` to `-n` in the other line to make it more read-able (as now it it fi this and that); and then my consistency check instinct told me to apply it here also.

lluiscampos:
The fix is not enough. I am working on a new one...

PR #79, opened 2020-06-03
[check_commits] Skip all checks for dependabot commits

lluiscampos:
Not only dependabot does not include Changelog tags, but also the commit
sign does not match the Git user.

This commit partially reverts incomplete fix from ffc2e5e and instead
just skips all checks for these commits.


lluiscampos:
Fix-up for #78

PR #80, opened 2020-06-03
[check_commits] Fix typo if conditions

lluiscampos:
Without quotes, `if [ -n ]` validates to true. Bug introduced at
ffc2e5e.

### DIFF
--- a/check_commits.sh
+++ b/check_commits.sh
@@ -95,7 +95,7 @@ do
         continue
     fi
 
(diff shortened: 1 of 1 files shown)


lluiscampos:
@oleorhagen You were right after all: I should not have touched that lines :)

lluiscampos:
Merging with no review: it is blocking PRs in other repositories.

PR #84, opened 2020-06-17
[check_commits] Accept "Changelog: All" and restore check for dependabot

lluiscampos:
* [check_commits] Accept "All" as a valid Changelog keyword
* [check_commits] Restore Changelog check for dependabot

### DIFF
--- a/check_commits.sh
+++ b/check_commits.sh
@@ -109,8 +109,8 @@ do
             echo >&2 "Commit ${i} doesn't have a changelog tag! Make a changelog entry for your commit (https://github.com/mendersoftware/mender/blob/master/CONTRIBUTING.md#changelog-tags)."
             notvalid="$notvalid $i"
         # Less than three words probably means something was misspelled, except for
-        # None, Title and Commit.
-        elif ! echo "$COMMIT_MSG" | egrep -i "^ *Changelog: *(None|Title|Commit|\S+(\s+\S+){2,}) *$" >/dev/null; then
+        # None, Title, Commit and All.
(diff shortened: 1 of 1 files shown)


check_commits.sh
```diff
@@ -109,8 +108,8 @@ do
             echo >&2 "Commit ${i} doesn't have a changelog tag! Make a changelog entry for your commit (https://github.com/mendersoftware/mender/blob/master/CONTRIBUTING.md#changelog-tags)."
             notvalid="$notvalid $i"
         # Less than three words probably means something was misspelled, except for
-        # None, Title and Commit.
-        elif ! echo "$COMMIT_MSG" | egrep -i "^ *Changelog: *(None|Title|Commit|\S+(\s+\S+){2,}) *$" >/dev/null; then
+        # None, Title, Commit and All.
+        elif ! echo "$COMMIT_MSG" | egrep -i "^ *Changelog: *(None|Title|Commit|All|\S+(\s+\S+){2,}) *$" >/dev/null; then
```
kacf, line 112: This will not match correctly for dependabot `Changelog: all` entries. The reason is that this checks whether `All` is the *only* word on the line, so when matching here it's actually reverting to the last branch (starting with `\S+`), which checks for number of words. So this is still subject to the error message below.

But maybe that's ok? I suppose that dependabot should always write at least a few words?

lluiscampos: @kacf Oh, I see what you mean. It will "pass" for this case as the dependabot messages are like "Changelog:All: Bump <component> from <version> to <version>" (see https://github.com/mendersoftware/mender-artifact/pull/280 for an example).

I already expent enough time with this task (namely, the other PR) so I think I will merge as-is.

kacf: I'm fine with that. Worst case we will get some false alarms.

PR #86, opened 2020-06-25
[docker-build-template] Fix publish:image:mender:dockerhub

lluiscampos:
Images :mender-X.X.x for enterprise repositories were not being
published into Docker Hub due to this bug in the template.


lluiscampos:
FYI @tranchitella

PR #87, opened 2020-06-29
QA-182: service pipeline: hanlding of saas-<VER> tags

tranchitella:


.gitlab-ci-check-docker-build.yml
```diff
@@ -191,3 +191,38 @@ trigger:saas:sync-staging-component:
...
+  only:
+    refs:
+      - /^saas-[a-zA-Z0-9.]+$/
+  tags:
+    - docker
+  image: docker
+  services:
+    - docker:19.03.5-dind
+  dependencies:
+    - build:docker
```
kacf, line 207: In the same pull request you are adding:

```diff
 build:docker:
   stage: build
+  except:
+    - /^saas-[a-zA-Z0-9.]+$/
```

Will that work? IIRC, jobs that have matching exceptions will just lead to no dependency, IOW the publish job will still run, but the build will not.

tranchitella: > I assume that this removal of testing is because all of this is expected to have gone through `staging` already?

Yes, the idea is that pushing a tag with the `saas-` prefix should only re-tag to already built and pushed docker image (staging_SHA). I don't know if this is going to work, I've tried my best but I don't know how to actually test the updated pipeline. @lluiscampos?

lluiscampos: @tranchitella Kristian is right in that this job cannot depend on `build:docker`. I should not have any dependency as this will just docker pull + docker tag + docker push (as you have correctly done).

.gitlab-ci-check-docker-build.yml
```diff
@@ -187,3 +191,38 @@ trigger:saas:sync-staging-component:
...
+  before_script:
+    - export SOURCE_TAG=staging_${CI_COMMIT_SHA}
+    - export DOCKER_TAG=${CI_COMMIT_REF_SLUG}
+    - export SERVICE_IMAGE=${DOCKER_REPOSITORY}:${DOCKER_TAG}
```
lluiscampos, line 211: To be consistent with the rest of the pipeline, I suggest renaming like:

```suggestion
    - export SOURCE_TAG=staging_${CI_COMMIT_SHA}
    - export DOCKER_PUBLISH_TAG=${CI_COMMIT_REF_NAME}
    - export SERVICE_IMAGE=${DOCKER_REPOSITORY}:${DOCKER_PUBLISH_TAG}
```

Using `CI_COMMIT_REF_SLUG` would work as we are planning the tags, but if we ever have incompatible tags, I rather let it fail and not use that tags.

.gitlab-ci-check-docker-build.yml
```diff
@@ -187,3 +191,38 @@ trigger:saas:sync-staging-component:
...
+# Extra job for Enterprise repos, publish also to Docker Hub
+publish:image:saas:dockerhub:
+  extends: publish:image:saas
+  only:
+    variables:
+      - $DOCKER_REPOSITORY =~ /^registry\.mender\.io.\/*/
+  before_script:
+    - export SOURCE_TAG=staging_${CI_COMMIT_SHA}
+    - export DOCKER_TAG=${CI_COMMIT_REF_SLUG}
+    - export SERVICE_IMAGE=${DOCKER_REPOSITORY#registry.mender.io/}:${DOCKER_TAG}
```
lluiscampos, line 228: If you apply my previous comment, fix here also.

lluiscampos:
@tranchitella  I forgot to say that you *almost* nailed it in the first try, so well done! :climbing_man:

.gitlab-ci-check-docker-build.yml
tranchitella: Ack.

.gitlab-ci-check-docker-build.yml
tranchitella: Done

tranchitella:
@lluiscampos thank you for your explanation, it is very valuable. I've added it to the KB:
https://gitlab.com/Northern.tech/Mender/knowledge-base/-/wikis/CI-CD#how-can-i-debug-a-gitlab-cicd-pipeline-for-example-the-saas-one

.gitlab-ci-check-docker-build.yml
tranchitella: Done!

PR #90, opened 2020-08-03
Add retries to acceptance/unit tests codecov upload

alfrunes:
It has happened a couple of times that the codecov script fails to
upload the coverage for either acceptance or unit tests, but rerunning
the stage succeeds.
*(This PR is just a suggested temporary solution to the problem)*

### DIFF
--- a/.gitlab-ci-check-docker-acceptance.yml
+++ b/.gitlab-ci-check-docker-acceptance.yml
@@ -108,4 +108,12 @@ publish:acceptance:
   before_script:
     - apk add --no-cache bash curl findutils git
   script:
-    - bash -c "bash <(curl -s https://codecov.io/bash) -Z -F acceptance -f ./tests/coverage-acceptance.txt"
+    - curl -s https://codecov.io/bash > /tmp/codecov.bash
+    - chmod +x /tmp/codecov.bash
+    - for i in $(seq 5); do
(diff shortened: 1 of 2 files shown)


.gitlab-ci-check-docker-acceptance.yml
```diff
@@ -108,4 +108,14 @@ publish:acceptance:
...
-    - bash -c "bash <(curl -s https://codecov.io/bash) -Z -F acceptance -f ./tests/coverage-acceptance.txt"
+    - curl -s https://codecov.io/bash > /tmp/codecov.bash
+    - chmod +x /tmp/codecov.bash
+    - retries=5
+    - while [ $retries -gt 0 ]; do
+        /tmp/codecov.bash -Z -F acceptance -f ./tests/coverage-acceptance.txt;
+        if [ $? -eq 0 ]; then
+          break;
+        fi
+        retries=$(expr $retries - 1);
```
lluiscampos, line 119: This is totally fine, but just for future reference: you can avoid the expr call by just doing:
```suggestion
    - for retries in 1 2 3 4 5; do
        /tmp/codecov.bash -Z -F acceptance -f ./tests/coverage-acceptance.txt;
        if [ $? -eq 0 ]; then
          break;
        fi
```

PR #95, opened 2020-08-26
fix: Make the 'check_license.sh' script fail on format errors

oleorhagen:
Previously, a misformed line, like a line with a sha that was missing a
character would pass the check, as the program would only produce a warning, and
not an error code.

This fix makes the program fail on such formatting errors.

Note that it does the verification on a temp-file, as the input license file
needs to be stripped of newlines, as these are also reported as formatting errors.

Example input:

```
8514c577c1c4b46cee454d5a882b15fa411e72c5bd7f801f241591789fce61a  vendor/github.com/spf13/pflag/LICENSE
f6918bd93ffe07f4b2c61b8287c32cb3122e08aed0be50f1c7d0eddc87877a8f  vendor/gopkg.in/ini.v1/LICENSE
5e3400b93bbb099e83e52bab885e7441750673c21f97988ca3f1240639b63283  vendor/github.com/spf13/afero/LICENSE.txt
b40930bbcf80744c86c46a12bc9da056641d722716c378f5659b9e555ef833e1  vendor/gopkg.in/yaml.v2/LICENSE

06d8d6fe9e7dca9cd1c6385eab60b67c34f59a7871ee794553904b0d5cfcd2c  vendor/github.com/howeyc/gopass/LICENSE.txt
1b93a317849ee09d3d7e4f1d20c2b78ddb230b4becb12d7c224c927b9d470251  vendor/github.com/davecgh/go-spew/LICENSE
```

Example output:

```
Some lines in the LIC_FILE_CHKSUM.sha256 file are misformed:
8514c577c1c4b46cee454d5a882b15fa411e72c5bd7f801f241591789fce61a  vendor/github.com/spf13/pflag/LICENSE
06d8d6fe9e7dca9cd1c6385eab60b67c34f59a7871ee794553904b0d5cfcd2c  vendor/github.com/howeyc/gopass/LICENSE.txt
```

Changelog: None

### DIFF
--- a/check_license.sh
+++ b/check_license.sh
@@ -48,8 +48,27 @@ then
     cd "$1"
 fi
 
-# Known licenses must continue to match.
-shasum -a 256 -c $CHKSUM_FILE
+# Remove all newlines from the Checksum file as these are reported as formatting
+# errors by the shasum program
+TMP_CHKSUM_FILE=$(mktemp)
+trap cleanup EXIT
+cleanup()
+{
+  rm $TMP_CHKSUM_FILE
+}
+sed '/^$/d' $CHKSUM_FILE > $TMP_CHKSUM_FILE
+
+# Use the tmp-file for the rest of the script
+CHKSUM_FILE=$TMP_CHKSUM_FILE
+
+# Collect only stderr from the subcommand
+output="$(exec 3>&1; shasum --warn --algorithm 256 --check $CHKSUM_FILE >/dev/null 2>&3)"
+
+if echo "$output" | grep -q 'line is improperly formatted' -; then
+  echo >&2 "Some line(s) in the LIC_FILE_CHKSUM.sha256 file are misformed"
+  cat $CHKSUM_FILE
+  exit 1
+fi
 
 # Unlisted licenses not allowed.
 for file in $(find . -iname 'LICEN[SC]E' -o -iname 'LICEN[SC]E.*' -o -iname 'COPYING')
--- a/license_test.go
+++ b/license_test.go
@@ -193,3 +193,38 @@ func TestCommercialLicense(t *testing.T) {
 	err = cmd.Run()
 	assert.NoError(t, err)
 }
+
+func TestMisformedLicenseChecksumLines(t *testing.T) {
+
+	t.Log("Testing Godep with license in README.md, with checksum")
+	// We need a custom LIC_FILES_CHKSUM.sha256, so use a temp dir
+	// for this one.
+	require.NoError(t, os.MkdirAll("tmp/vendor/dummy-site.org/test-repo", 0755))
+	require.NoError(t, os.Chdir("tmp"))
+	defer os.RemoveAll("tmp")
+	defer os.Chdir("..")
+	fd, err := os.Create("vendor/dummy-site.org/test-repo/test.go")
+	require.NoError(t, err)
+	fd.Close()
+	fd, err = os.Create("vendor/dummy-site.org/test-repo/README.md")
+	require.NoError(t, err)
+	fd.Close()
+	fd, err = os.Create("LICENSE")
+	fmt.Fprintln(fd, "Copyright 2020 Northern.tech")
+	require.NoError(t, err)
+	fd.Close()
+
+	fd, err = os.Create("LIC_FILES_CHKSUM.sha256")
+	// This is one letter short of a full shasum line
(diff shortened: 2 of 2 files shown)


check_license.sh
```diff
@@ -48,8 +48,37 @@ then
...
+# Collect only stderr from the subcommand
+output="$(exec 3>&1; shasum --warn --algorithm 256 --check $CHKSUM_FILE >/dev/null 2>&3)"
+
+if ! grep -q 'lines are improperly formatted' 2>/dev/null < "$output"; then
+  echo >&2 "Some lines in the LIC_FILE_CHKSUM.sha256 file are misformed:"
+  (
+    IFS="
+"
+    for line in $output; do
+      if [ "$line" = "shasum: WARNING: 2 lines are improperly formatted" ]; then
```
lluiscampos, line 73: This `2` here looks suspicious...

check_license.sh
```diff
@@ -48,8 +48,37 @@ then
...
+CHKSUM_FILE=$TMP_CHKSUM_FILE
+
+# Collect only stderr from the subcommand
+output="$(exec 3>&1; shasum --warn --algorithm 256 --check $CHKSUM_FILE >/dev/null 2>&3)"
+
+if ! grep -q 'lines are improperly formatted' 2>/dev/null < "$output"; then
+  echo >&2 "Some lines in the LIC_FILE_CHKSUM.sha256 file are misformed:"
+  (
+    IFS="
+"
```
lluiscampos, line 71: Can you help me understanding why we need this?

check_license.sh
```diff
@@ -48,8 +48,37 @@ then
...
+{
+  rm $TMP_CHKSUM_FILE
+}
+sed '/^$/d' $CHKSUM_FILE > $TMP_CHKSUM_FILE
+
+# Use the tmp-file for the rest of the script
+CHKSUM_FILE=$TMP_CHKSUM_FILE
+
+# Collect only stderr from the subcommand
+output="$(exec 3>&1; shasum --warn --algorithm 256 --check $CHKSUM_FILE >/dev/null 2>&3)"
```
lluiscampos, line 65: Nice trick. Noted.

check_license.sh
```diff
@@ -48,8 +48,37 @@ then
...
+  echo >&2 "Some lines in the LIC_FILE_CHKSUM.sha256 file are misformed:"
+  (
+    IFS="
+"
+    for line in $output; do
+      if [ "$line" = "shasum: WARNING: 2 lines are improperly formatted" ]; then
+        continue
+      fi
+      linenr="$(echo $line | sed -E 's;shasum: /tmp/tmp\S+: ([0-9]+).*;\1;')"
+      awk "NR == $linenr " $CHKSUM_FILE
```
lluiscampos, line 77: I _think_ I understand the logic and the purpose of this. But I think is an overkill :)

TBH I don't think it is necessary to process the output so much as we risk hiding the error if the output of  `shasum` is slightly different. On failures I would just print the full `CHKSUM_FILE` as it will be quite obvious for the PR author to see what has he/she messed up.

As you have already done "the job" I don't have objections on leaving it; I just wanted to share my point of view for future reference.

lluiscampos (changes requested):
Thanks for following up. Well done!

From my comments, the only thing we must fix (or explain!) is the hard coding of `2`. The rest are clarifications or suggestions.

check_license.sh
oleorhagen: Yes, I'm on atm. This is an error.

check_license.sh
oleorhagen: It's just for word splitting in the following for loop to only split on newlines

check_license.sh
oleorhagen: Alright. I agree. Let's keep things simple.

lluiscampos: KISS

lluiscampos:
You probably realized, but we need to fix the unit tests also.

check_license.sh
lluiscampos: Ah, of course. Thanks

oleorhagen:
Yup, on it. They actually picked up on something also

oleorhagen:
@lluiscampos 
Simplified the error output significantly. It now prints an error, and outputs the file in whole.
Fixed the errors.
Also added a test for this specific case.

lluiscampos (approved):

PR #97, opened 2020-09-08
MEN-2917: upgrade mongodb to 4.4 and golang to 1.14

tranchitella:


.gitlab-ci-check-golang-unittests.yml
```diff
@@ -15,7 +15,7 @@
 # The template uses latest Mender supported golang version.
 # To override the version to use, append to your .gitlab-ci.yml:
 # test:unit:
-#   image: golang:1.11.4
+#   image: 1.11.4
```
lluiscampos, line 18: This should still contain "golang"

.gitlab-ci-check-golang-unittests.yml
```diff
@@ -40,10 +40,10 @@ test:unit:
     - GO111MODULE=off go get -u golang.org/x/tools/cmd/cover
 
     # Install mongodb for the tests that use it locally
-    - apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv 2930ADAE8CAF5059EE73BB4B58712A2291FA4AD5
-    - echo "deb http://repo.mongodb.org/apt/debian stretch/mongodb-org/3.6 main" | tee /etc/apt/sources.list.d/mongodb-org-3.6.list
+    - wget -qO - https://www.mongodb.org/static/pgp/server-4.4.asc | sudo apt-key add -
+    - echo "deb http://repo.mongodb.org/apt/debian buster/mongodb-org/4.4 main" | tee /etc/apt/sources.list.d/mongodb-org-4.4.list
     - apt-get -qq update
-    - apt-get install -qy --allow-unauthenticated mongodb-org-server=3.6.11
+    - apt-get install -qy mongodb-org
```
lluiscampos, line 47: The challenge here is that we don't know if it will work for all the repos before merging...

@tranchitella @mzedel @kacf I guess the time has arrived to figure out a procedure on updating the templates! Let's freeze this PR for couple of days and discuss ideas.

My suggestion:
1. Merge this into `master-next` branch or similar
2. Launch PRs in all repos (master and supported release branches) modifying the include to master-next. This can be done with a little bit of scripting.
3. If all passes, close the above PRs, merge `master-next` to `master` in here.

In any case, maybe we should have a JIRA ticket for it, so that it gets documented for the next update.

lluiscampos (changes requested):

.gitlab-ci-check-golang-unittests.yml
tranchitella: sorry, I replaced it by mistake

.gitlab-ci-check-golang-unittests.yml
tranchitella: @lluiscampos do you mean if all the repos are ready to run unit tests using mongodb 4.4? I've tested it locally (back-end) and they are okay.

mzedel: In case of a larger change in the way the pipelines work, I would think the process you suggested is the way to go. With version/ dependency upgrades as in here however, letting things fail would highlight either areas we have to fix soon anyway (as the mongo version in integration is what we release, or the golang version is what we have in our `Dockerfile`s) or point to areas in the tests, that might be version specific and should thus be improved. During a release this might not be advisable but in the "off-season" this should increase the urgency to fix what we have in master...

tranchitella: thanks @mzedel, this is exactly the reason why I would like to run unit tests with the same versions we released for (mongo and golang). If we have issues, I can fix them straightaway, better than not knowing it at all, IMHO.

lluiscampos: Agreed in the procedure. Let's just do it in couple of days after the 2.5.0 release is built.

Keep in mind that I said "releases" in plural because we need to check all maintained releases (usually two), not only the latest one.

lluiscampos (approved):
Approved, we'll merge in couple of days time.

.gitlab-ci-check-golang-unittests.yml
tranchitella: Ack

tranchitella:
@lluiscampos can we merge this PR?

lluiscampos:
@tranchitella Yes, go ahead :+1: 

FYI @kjaskiewiczz, some individual pipelines might start failing after this.

PR #98, opened 2020-09-12
QA-220: Migrate coverage reporting to coveralls

lluiscampos:
This PR brings coveralls integration without removing codecov, so that the development process does not get affected for now.

For now, only `devicauth` is activated and you can see the it working in [this PR](https://github.com/mendersoftware/deviceauth/pull/397) and in [coveralls](https://coveralls.io/github/mendersoftware/deviceauth).

I need however to merge these changes into master to be able to finish the PoC; mainly because until we don't have reports from master we cannot see how the "diff" feature works.

PR #99, opened 2020-09-14
Revert "QA-220: Migrate coverage reporting to coveralls"

lluiscampos:
Reverts mendersoftware/mendertesting#98


lluiscampos:
@oleorhagen Reverting, as I just realized that this will break absolutely all repositories that are not activated in coveralls (and we cannot active the close source ones until NT approves the expense).

PR #103, opened 2020-09-17
[apidocs template] Use BUILD_X_Y instead of LATEST/OLDSTABLE

lluiscampos:
The problem with using LATEST/OLDSTABLE is that it was hardcoded in the
mender-api-docs repository, so that for a short while (while the release
is in progress) we would actually be building latest (say, 2.5) while
publishing as oldstable (say, 2.4).

Note that this has not been a problem because as per today we have not
had 2.4/2.3 API docs in production.

A more permanent fix will follow.


lluiscampos:
See also https://github.com/mendersoftware/mender-api-docs/pull/22

tranchitellawfp (approved):

tranchitella (approved):

PR #105, opened 2020-09-28
[unittests template] Update doc to name coveralls.io

lluiscampos:


lluiscampos:
Thanks @merlin-northern for reporting it

PR #109, opened 2020-10-14
[check-apidocs pipepeline] Support manual trigger for API site build

lluiscampos:
So that the API site can be rebuild without changes to docs/* files,
which might come in handy during some maintenance tasks.

To support that, the limited only/except logic has been replaced with
the more expressive rules syntax.


oleorhagen (approved):
Nice!

PR #116, opened 2020-11-13
Enhance gitlab-ci-check-apidocs to make it generalized

tranchitella:
The trigger:apidocs:rebuild-mender-api-docs job is now generalized and
does not require customization anymore for Mender specific versions.
The job relies entirely on the release_tool to determine which version
needs rebuilding, and sets the BUILD_X_Y=true variable accordingly when
triggering the rebuild of the API docs repository.

Changelog: none

Signed-off-by: Fabio Tranchitella <fabio.tranchitella@northern.tech>


lluiscampos (approved):
Independently of further improvements, this looks just great :top:

PR #123, opened 2020-12-21
Add(check_license.sh): Check license headers in Go and Python

oleorhagen:
check_license.sh now also checks for the license headers in all python and go source files.

This is handled by a new `check_license_headers.sh` file.


check_license.sh
```diff
@@ -71,65 +81,86 @@ if echo "$output" | grep -q 'line is improperly formatted' -; then
...
 do
     file=$(echo $file | sed -e 's,./,,')
     if ! fgrep "$(shasum -a 256 $file)" $CHKSUM_FILE > /dev/null
     then
-        echo "$file has missing or wrong entry in $CHKSUM_FILE"
+        echo >&2 "$file has missing or wrong entry in $CHKSUM_FILE"
         ret=1
     fi
-done
+done < <(find . -iname 'LICEN[SC]E' -o -iname 'LICEN[SC]E.*' -o -iname 'COPYING' | grep -v venv)
```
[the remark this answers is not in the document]
oleorhagen: I agree actually. I was of the opposite mind yesterday though

check_license.sh
```diff
@@ -71,65 +81,86 @@ if echo "$output" | grep -q 'line is improperly formatted' -; then
...
+SOURCE_FILES="\
+$(find . -type f \( ! -regex ${LICENSE_HEADERS_IGNORE_FILES_REGEXP} ! -path './vendor/*' -name '*.go' \))
+$(find . -type f \( ! -regex ${LICENSE_HEADERS_IGNORE_FILES_REGEXP} ! -regex '.*\.venv.*' ! -regex '.*build/.*' -name '*.py' \))"
+
+for source_file in ${SOURCE_FILES}; do
+  case ${source_file} in
+    *.go) CM='//' ;;
+    *.py) CM='#' ;;
+  esac
+  ~/mendersoftware/mendertesting/check_license_header.sh --comment-syntax "${CM}" "${source_file}" || { ret=1; true; }
```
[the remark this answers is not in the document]
oleorhagen: Yes! Agree. I came to this realization on the way home too.
Also, the path is leftover from local testing. Nearly got me out of bed last night

oleorhagen:
@kacf you can have a look now.

I'm not sure how/if we should handle script with starting shebangs or not really :thinking: 

Also, this is bound to break a lot of projects, due to the requirement for the negative filter when aggregating all the source files.

check_license_source_files.sh
```diff
@@ -0,0 +1,202 @@
...
+            ENT_COMMIT="${1#--ent-start-commit=}"
+            ;;
+        --ent-start-commit)
+            shift
+            ENT_COMMIT="$1"
+            ;;
+        --verbose)
+          set -x
+          VERBOSE="--verbose"
+          ;;
```
kacf, line 48: Nitpick: Wrong indentation.

licenses.go
```diff
@@ -67,7 +67,7 @@ func checkMenderCompliance() error {
 		return err
 	}
 
-	args := []string{path.Join(pathToTool, "check_license_go_code.sh")}
+	args := []string{path.Join(pathToTool, "check_license_source_files.sh")}
```
kacf, line 70: Should the old file be removed now?

check_license_source_files.sh
```diff
@@ -0,0 +1,202 @@
+#!/bin/bash
+
+# This regular expression can be set in the '.gitlab-ci.yml' file, and is passed
```
kacf, line 3: This file doesn't have `set -e`. It won't actually fail when there is a license problem.

check_license_source_files.sh
```diff
@@ -0,0 +1,202 @@
...
+        case ${source_file} in
+          *.go)
+              CM='//'
+              ;;
+          *.py)
+              CM="#"
+              ;;
+          *)
+              echo >&2 "No source file testing done for file:\n\t ${source_file}\nThe filetype is unsupported type"
+      esac
```
kacf, line 182: Isn't `;;` missing here? I thought bash wouldn't accept this but, maybe I'm wrong.

check_license_source_files.sh
```diff
@@ -0,0 +1,202 @@
...
+    for source_file in ${SOURCE_FILES}; do
+        case ${source_file} in
+          *.go)
+              CM='//'
+              ;;
+          *.py)
+              CM="#"
+              ;;
+          *)
+              echo >&2 "No source file testing done for file:\n\t ${source_file}\nThe filetype is unsupported type"
```
kacf, line 181: This just continues to `check_file` below with `CM` unset. I would suggest setting `TEST_RESULT=1` and calling `continue` here, so that it loops around and we catch it.

kacf:
> Also, this is bound to break a lot of projects, due to the requirement for the negative filter when aggregating all the source files.
> 
> The breaking test will always file, since it is trying to use the old filename: `check_license_go_source.sh`, which is no longer present (the template is taken from master).

I didn't understand this...

licenses.go
oleorhagen: Yes, and it is, no?

oleorhagen:
> > Also, this is bound to break a lot of projects, due to the requirement for the negative filter when aggregating all the source files.
> > The breaking test will always file, since it is trying to use the old filename: `check_license_go_source.sh`, which is no longer present (the template is taken from master).
> 
> I didn't understand this...

I'm sorry, let my try to rephrase:
The `.gitlab-ci-check-license.yml` file, which is included by `.gitlab-ci.yml` file, now tries to run `check_license_source_files.sh` from master, as far as I can see, due to the job:
```
test:check-license:
  stage: test
  needs: []
  image: alpine
  variables:
    GIT_DEPTH: 0
  before_script:
    # Install dependencies
    - apk add --no-cache git bash perl-utils
    # Rename the branch we're on, so that it's not in the way for the
    # subsequent fetch. It's ok if this fails, it just means we're not on any
    # branch.
    - git branch -m temp-branch || true
    # Git trick: Fetch directly into our local branches instead of remote
    # branches.
    - git fetch origin 'refs/heads/*:refs/heads/*'
    # Get last remaining tags, if any.
    - git fetch --tags origin
    - git clone --depth=1 http://github.com/mendersoftware/mendertesting /tmp/mendertesting
  script:
    # Check licenses
    - /tmp/mendertesting/check_license.sh
    - /tmp/mendertesting/check_license_source_files.sh
```
But this file is still missing in master

licenses.go
kacf: Lol, yes, it is! Github fooled me, I was looking for the "negative diff", but of course it is hidden behind "Load diff".

lluiscampos (commented):
I didn't look into the details (Kristian did already that).

> Also, this is bound to break a lot of projects, due to the requirement for the negative filter when aggregating all the source files.

I am okey with this. Let's do it soon so that we can fix the repos _before_ the release. Me being the build master this week I can coordinate it.

lluiscampos (approved):

oleorhagen:
Alright, can one of you guys merge then?

lluiscampos:
> Alright, can one of you guys merge then?

I can merge tomorrow morning and monitor all the pipelines.

PR #132, opened 2021-01-19
go fmt outputs empty line, sometimes.

merlin-northern:
example job:
https://gitlab.com/Northern.tech/Mender/go-lib-micro/-/jobs/971381446

from this pr:
https://github.com/mendersoftware/go-lib-micro/pull/100

and indeed it happens:
```
#echo "$(go fmt ./...)"

```

ChangeLog:none

### DIFF
--- a/.gitlab-ci-check-golang-static.yml
+++ b/.gitlab-ci-check-golang-static.yml
@@ -40,7 +40,7 @@ test:static:
     # Test if code was formatted with 'go fmt'
     # Command will format code and return modified files
     # fail if any have been modified.
-    - if [ -n "$(go fmt ./...)" ]; then echo 'Code is not formatted with "go fmt"'; false; fi
+    - if [ -n "$(go fmt ./... | grep -v ^$)" ]; then echo 'Code is not formatted with "go fmt"'; false; fi
     # Perform static code analysys
     - go vet `go list ./... | grep -v vendor`
     # Fail builds when the cyclomatic complexity reaches 20 or more


.gitlab-ci-check-golang-static.yml
```diff
@@ -40,7 +40,7 @@ test:static:
     # Test if code was formatted with 'go fmt'
     # Command will format code and return modified files
     # fail if any have been modified.
-    - if [ -n "$(go fmt ./...)" ]; then echo 'Code is not formatted with "go fmt"'; false; fi
+    - if [ -n "$(go fmt ./... | grep -v ^$)" ]; then echo 'Code is not formatted with "go fmt"'; false; fi
```
kacf, line 43: I think you need to single-quote `^$`. I'm not sure when the `$` is at the end like that, but I think it's a safer approach anyway.

lluiscampos (commented):
I am not sure I understand the consequences of this.

I think this will make the test ignore new lines, but then a local run of `go fmt ./...` could also add these lines back, right? The test will pass the next round also, but why don't be strict which these empty lines in the first place?

Also, if we merge this, @merlin-northern please follow up on pipelines for `mender`, `mender-artifact`, `mender-connect` (possibly others?).

alfrunes:
@merlin-northern ping

merlin-northern:
> @merlin-northern ping

wow. I am putting it on my todolist. I will come back to it.

merlin-northern:
> @merlin-northern ping

wow. I am putting it on my todolist. I will come back to it.

lluiscampos:
Doing the job of `stalebot`... This PR is now too old. Closing.

PR #134, opened 2021-01-29
[docker-build template] Do not fail on missing credentials

lluiscampos:
So that the template can be used on repositories with credentials for a
single registry.


tranchitella (approved):

0lmi:
@lluiscampos this is great improvement 👍

PR #138, opened 2021-03-05
 [check-docker-*] Replace only/except with modern rules (+ cleanup)

lluiscampos:
* [check-docker-*] Replace only/except with modern rules
The main motivation is to have an easier way to override rules by user
pipelines.
According to GitLab documentation only/except are not going to be
developed further, so better to switch to the new standard.

* [check-docker-build] Deprecate publishing :M.N.x tags
Since Mender 2.3 went EOL, there is no more use for these tags (we now
publish release branches using :mender-M.N.x tags)

### DIFF
--- a/.gitlab-ci-check-docker-acceptance.yml
+++ b/.gitlab-ci-check-docker-acceptance.yml
@@ -35,8 +35,9 @@ stages:
 test:prepare_acceptance:
   stage: test_prep
   image: docker
-  except:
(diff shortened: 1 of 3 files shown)


lluiscampos:
@alfrunes I think you are knowledgeable when it comes to these `rules` syntax. Can you please review carefully?

tranchitella (approved):

PR #140, opened 2021-03-08
[check-docker-*] Amend rules

lluiscampos:
When no match, rules default to not adding the job in the pipeline. So
all the jobs that were excluded for saas tags require a final catch all
rule for them to be included in the pipelines.

Amends commit 431bc37f452d155ef555852315ba947f67fac1c5


lluiscampos:
See https://docs.gitlab.com/ee/ci/yaml/README.html#rules-clauses, specifically the second example there.

tranchitella (approved):

oleorhagen (approved):

PR #145, opened 2021-04-26
[docker-acceptance] Improve support on repos using common tester image

lluiscampos:
So far the repos that have been migrated have kept a dummy Dockerfile
just to satisfy the template. Just check for the file here instead.

Extra:
* [docker-acceptance] Collect results as GitLab junit report

### DIFF
--- a/.gitlab-ci-check-docker-acceptance.yml
+++ b/.gitlab-ci-check-docker-acceptance.yml
@@ -101,6 +101,8 @@ test:acceptance_tests:
     paths:
       - tests/acceptance.*
       - tests/coverage-acceptance.txt
+    reports:
+      junit: tests/results.xml
     when: always
 
 publish:acceptance:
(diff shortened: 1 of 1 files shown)


.gitlab-ci-check-docker-acceptance.yml
```diff
@@ -101,6 +106,8 @@ test:acceptance_tests:
     paths:
       - tests/acceptance.*
       - tests/coverage-acceptance.txt
+    reports:
+      junit: tests/results.xml
```
mzedel, line 110: will this be generated by the common tester image?

.gitlab-ci-check-docker-acceptance.yml
```diff
@@ -51,15 +51,17 @@ test:prepare_acceptance:
...
+    -   docker save testing > acceptance_testing_image.tar
+    - fi
     - wget https://d1b0l86ne08fsf.cloudfront.net/mender-artifact/master/linux/mender-artifact
     - chmod +x mender-artifact
   artifacts:
     expire_in: 2w
     paths:
-      - tests_image.tar
-      - acceptance_testing_image.tar
+      - '*_image.tar'
```
mzedel, line 64: nice 😄

mzedel (approved):
👍 nice idea - I appreciate the related cleanups!

.gitlab-ci-check-docker-acceptance.yml
lluiscampos: Not the image itself, but the [wrapper script](https://github.com/mendersoftware/deviceauth/blob/master/tests/run.sh#L21) in each repository. I *think* CI won't fail the job is the file is missing, it will just not parse it.

You can see it working [here](https://gitlab.com/Northern.tech/Mender/deviceauth/-/pipelines/292362776/test_report).

mzedel: thanks for the pointers!

PR #148, opened 2021-05-11
[pipeline] Remove codecov integration

lluiscampos:
This go code is not in use anyway, and seems like this is the only repo
still using codecov.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -5,7 +5,6 @@ include:
 
 stages:
   - test
-  - publish
 
 # See QA-274
 # test:lint:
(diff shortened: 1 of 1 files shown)


mzedel (approved):
Thanks for cleaning that up! Could we also get rid of the coverage.txt/ the coverage collection step in the unit tests - or is this useful to have in some instances?

lluiscampos:
> Thanks for cleaning that up! Could we also get rid of the coverage.txt/ the coverage collection step in the unit tests - or is this useful to have in some instances?

We could... but I thought it didn't hurt neither.

As I have noted in the past, the go code is not used, is here only to have some tests on these bash scripts. We could think of moving the test code to Python and get the rid of all the go code.

PR #149, opened 2021-05-28
Add template to check shell formatting with shfmt

lluiscampos:


lluiscampos:
Based on https://github.com/mendersoftware/mender-convert/blob/12d9af29a8a270b5303327f166683e6c120ccdfa/.gitlab-ci.yml#L56

lluiscampos:
Tested "successfully" in [this pipeline](https://gitlab.com/Northern.tech/Mender/mender-convert/-/jobs/1301920221).

.gitlab-ci-check-shell-format.yml
```diff
@@ -0,0 +1,72 @@
...
+  image: alpine:3.13
+  before_script:
+    - SHELL_SCRIPTS=$(find . -type f -name "*.sh")
+  script:
+    - apk update
+    - apk upgrade
+    - apk add --update bash git shfmt
+    - shfmt -version
+    - /bin/bash
+    - 'echo -e "Checking shell formatting for the following files:\n${SHELL_SCRIPTS}"'
```
merlin-northern, line 40: is this in single quotes on purpose?

merlin-northern (approved):
thank you!

kacf (approved):

.gitlab-ci-check-shell-format.yml
```diff
@@ -0,0 +1,72 @@
...
+
+stages:
+  - test
+
+test:check-shell-formatting:
+  stage: test
+  needs: []
+  image: alpine:3.13
+  before_script:
+    - SHELL_SCRIPTS=$(find . -type f -name "*.sh")
```
kacf, line 33: I suppose you could also find scripts whose first line is `#!/bin/(ba)?sh`.

oleorhagen: Good idea!
Let's do both

oleorhagen (approved):
Real nice!

.gitlab-ci-check-shell-format.yml
lluiscampos: The single quotes is to make the line correctly parsed as a yaml entry (due to the `:` half way)

.gitlab-ci-check-shell-format.yml
lluiscampos: We could, but then we should also accept `#! /usr/bin/env bash` (used in mender-convert) and we would potentially be missing files that are meant to be sourced.

> Let's do both

This is a bad idea from a template point of view, I think. A template should define "an standard" (for example based on file extension, or based on the shebang) but not both. Then a non-standard project can do whatever they want defining this variable.

lluiscampos: Thanks for your suggestions but I'll leave it as-is. Basically because I already have it working and tested :wrench:

PR #151, opened 2021-06-04
[docker templates] Revert to "dependencies" for publishing jobs

lluiscampos:
This partially reverts 09f6522e, resulting a mix of "needs" for the
build job and "dependencies" for publish jobs. The motivation is that we
shall not publish containers if there is any job failing in a previous
stage.

Reasoning can be found at:
https://github.com/mendersoftware/gui/pull/1603#issuecomment-843841981

Changelog: None


mzedel (approved):
🙏 I really appreciate that!

PR #152, opened 2021-06-17
FIX: coveralls not reporting back coverage to branches with pr_ prefix

alfrunes:


.gitlab-ci-check-docker-acceptance.yml
```diff
@@ -135,11 +135,10 @@ publish:acceptance:
     #  According to https://docs.coveralls.io/supported-ci-services
     #  we should set CI_NAME, CI_BUILD_NUMBER, etc. But according
     #  to goveralls source code (https://github.com/mattn/goveralls)
-    #  many of these are not supported. Set CI_BRANCH, CI_PR_NUMBER,
+    #  many of these are not supported. Set CI_BRANCH
     #  and pass few others as command line arguments.
     #  See also https://docs.coveralls.io/api-reference
     - export CI_BRANCH=${CI_COMMIT_BRANCH}
-    - export CI_PR_NUMBER=${CI_COMMIT_BRANCH#pr_}
```
alfrunes, line 142: FYI; this env variable is not required according to the API specification, and also breaks the API if it is not a `pr_*` branch (which is why I think the coverage is not reported).

lluiscampos: Okey, thanks. Maybe something has changed in either `goveralls` or the Coveralls API during this year, because I remember this being necessary to be exported when I originally made the migration.

lluiscampos (approved):
Thanks for looking into this. Have you tested it?

alfrunes:
> Thanks for looking into this. Have you tested it?

No, but I will create a dummy PR and test it as soon as I hit 'merge'. I reviewed the mattn/goveralls source code and it seems to always be filling out the `git` parameter, which according to the api reference should be enough for coveralls update the PR status.

lluiscampos:
Go ahead with the merge. No objection from my side.

PR #153, opened 2021-06-22
check_license.sh: Print shasum output when mismatches

lluiscampos:
On checksum mismatch shasum returns non-zero and the script exists
silently. Now it will print the warning before clean-up.
For example: "shasum: WARNING: 1 computed checksum did NOT match"

### DIFF
--- a/check_license.sh
+++ b/check_license.sh
@@ -54,6 +54,9 @@ TMP_CHKSUM_FILE=$(mktemp)
 trap cleanup EXIT
 cleanup()
 {
+  if [ -n "$output" ]; then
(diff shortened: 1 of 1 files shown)


kacf (approved):

kacf:
Btw, the indentation in this file is inconsistent. What you did is correct for that location, but to follow our normal style it should be four spaces.

lluiscampos:
> Btw, the indentation in this file is inconsistent. What you did is correct for that location, but to follow our normal style it should be four spaces.

I have run `shfmt` on the script :+1:

PR #158, opened 2021-09-29
Add mender-ci-common.sh with helper functions for any pipeline

lluiscampos:
Implementing save/load from S3 temp bucket.

The idea is to be able to extend this file with any function that might
be shared by two ore more pipelines.

### DIFF
--- a/mender-ci-common.sh
+++ b/mender-ci-common.sh
@@ -1,0 +1,60 @@
+# mender-ci-common.sh
+#
+# This file defines shell helper functions to be used by any Mender pipeline.
+# It is meant to be sourced directly from the GitLab CI jobs. For example:
+#
+#
+# my-fantastic-job:
+#   before_script:
+#     - eval "$(curl https://raw.githubusercontent.com/mendersoftware/mendertesting/master/mender-ci-common.sh)"
+#
+# Read the documentation for each function for more details
+#
+
+# Usage: mender_ci_save_tmp_artifact file-to-save
+#
+# Saves the given file in S3 bucket temporary storage.
+# For security, it computes the SHA256 checksum and saves it in
+# ${CI_PROJECT_DIR}/checksums. It is responsibility of the caller to save the
+# directory as part of the CI job artifacts:
+#   artifacts:
+#     paths:
+#       - checksums
+# Later jobs that will call mender_ci_load_tmp_artifact have to define "needs:"
+# or "depends:" to load the checksums directory
+function mender_ci_save_tmp_artifact {
+    if [ -z "$1" ]; then
+        echo Usage $0 file-to-save [project-name] [pipeline-id]
+        exit 1
+    fi
+    user_file="$1"
+    project_name=${2-$CI_PROJECT_NAME}
+    pipeline_id=${3-$CI_PIPELINE_ID}
+    if [ -z "$project_name" -o -z "$pipeline_id" ]; then
+        echo "Could not get project-name or pipeline-id. Either set env variables or pass them as parameters"
+        exit 1
+    fi
+    mkdir -p ${CI_PROJECT_DIR}/checksums
+    sha256sum ${user_file} > ${CI_PROJECT_DIR}/checksums/$(basename ${user_file}).sha256
+    aws s3 cp ${user_file} s3://mender-gitlab-tmp-storage/$project_name/$pipeline_id/$(basename ${user_file})
+}
+
+# Usage: mender_ci_load_tmp_artifact file-to-load
+#
+# Loads the given file from S3 bucket temporary storage. It checks the SHA256
+# checksum (see mender_ci_save_tmp_artifact).
+function mender_ci_load_tmp_artifact {
+    if [ -z "$1" ]; then
+        echo Usage $0 file-to-save [project-name] [pipeline-id]
+        exit 1
+    fi
+    user_file="$1"
+    project_name=${2-$CI_PROJECT_NAME}
+    pipeline_id=${3-$CI_PIPELINE_ID}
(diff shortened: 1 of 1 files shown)


lluiscampos:
@mzedel @kacf @oleorhagen What do you think about this? You can see a tentatively usage in `mender-convert`:
* https://github.com/mendersoftware/mender-convert/pull/390

Although the real motivation is to reuse this code in `mender-dist-packages`:
* https://github.com/mendersoftware/mender-dist-packages/pull/134

And possibly other places! @mzedel Extending these functions a bit we could also share Artifacts between different pipelines, like `mender-qa` and `mender-convert`, for example.

lluiscampos:
Okey, so this doesn't work as expected. See issues https://gitlab.com/gitlab-org/gitlab/-/issues/5810 and https://gitlab.com/gitlab-org/gitlab/-/issues/5078

Anyway, the idea is still valid. We can instead have here a `.sh` file meant to be sourced. I'll wait for feedback before reworking.

.gitlab-ci-common.yml
```diff
@@ -0,0 +1,52 @@
...
+
+# Usage: mender_ci_load_tmp_artifact file-to-load
+#
+# Loads the given file from S3 bucket temporary storage. It checks the SHA256
+# checksum (see mender_ci_save_tmp_artifact).
+.mender_ci_load_tmp_artifact: &mender_ci_load_tmp_artifact |
+  function mender_ci_load_tmp_artifact {
+    if [ -z "$1" ] then; echo Usage $0 file-to-load; exit 1; fi
+    user_file="$1"
+    aws s3 cp s3://mender-gitlab-tmp-storage/$CI_PROJECT_NAME/$CI_PIPELINE_ID/$(basename ${user_file}) ${user_file}
```
mzedel, line 50: I think the `PROJECT_NAME` and `PIPELINE_ID` need to also be parameters, as this will differ in (downstream) pipelines...

.gitlab-ci-common.yml
```diff
@@ -0,0 +1,52 @@
...
+    sha256sum -c ${CI_PROJECT_DIR}/checksums/$(basename ${user_file}).sha256
```
mzedel, line 51: that might be useful to also get from s3 in order to compare it here

mzedel (commented):
That should become pretty handy to have... Because of the `aws` dependency I doubt however that this would work as flexible as described e.g. if it is not available in the job you want to have the files in (as the package might not be available on `fancy-distribution-xy` or `alpine` I think)...
However as a sourced script in a job that runs just if an upstream `PROJECT_NAME` & `PIPELINE_ID` are set 👌...

.gitlab-ci-common.yml
lluiscampos: For security reasons, we need to have it in the GitLab pipeline. What this is trying to protect is against someone running a different pipeline and tampering your files, as the keys for this temporary bucket are not protected. See https://github.com/mendersoftware/mender-convert/commit/9d18a8717d8812dba135c3e2231e786afeb1ac6c

.gitlab-ci-common.yml
lluiscampos: These functions are meant to be used in the same pipeline. Mimic what today we have in mender-convert and I want to reuse in mender-dist-packages.

To cover the use case you are referring to (to share between pipelines) I would implement different functions, where we can either get PROJECT_NAME  and PIPELINE_ID as parameters or maybe just generate random dir and return the name to the caller or something. We can think about it when the time comes for implementing such case.

.gitlab-ci-common.yml
mzedel: oh ok, so I guess then the checksum comparison is left to the outside pipeline, I thought this was to be contained in these functions... (besides also expecting this to be used across pipelines)

.gitlab-ci-common.yml
mzedel: IMO this is close enough to the share across pipelines functionality that it might be worth supporting that... apart from that I would still recommend explicit parameters instead of relying on global variables (also as a general programming practice).

.gitlab-ci-common.yml
lluiscampos: It is tricky, but is the only way we think it can be safe. The checksum comparison is done by these functions (note `sha256sum -c`), the only left to the outside pipeline is to collect the `checksums/` as job artifacts.

.gitlab-ci-common.yml
lluiscampos: Okey, you convinced me. I'll rework :)

lluiscampos:
Converted into a shell script file and reworked following @mzedel feedback.

mender-ci-common.sh
```diff
@@ -0,0 +1,61 @@
...
+  project_name=${2-$CI_PROJECT_NAME}
+  pipeline_id=${3-$CI_PIPELINE_ID}
+  if [ -z "$project_name" -o -z "$pipeline_id" ]; then
+    echo "Could not get project-name or pipeline-id. Either set env variables or pass them as parameters"
+    exit 1
+  fi
+  mkdir -p ${CI_PROJECT_DIR}/checksums
+  sha256sum ${user_file} > ${CI_PROJECT_DIR}/checksums/$(basename ${user_file}).sha256
+  aws s3 cp ${user_file} s3://mender-gitlab-tmp-storage/$project_name/$pipeline_id/$(basename ${user_file})
+}
```
kacf, line 41: We agreed on four-space indented shell code some time ago (mostly because of Yocto). :-)

kacf (approved):
Very much in favor!

oleorhagen (approved):
Cool idea!

PR #159, opened 2021-10-04
mender-ci-common.sh: Use POSIX syntax

lluiscampos:
So that the functions can be sourced by non-bash shells.


lluiscampos:
A follow-up from #158.

oleorhagen (approved):
I'm guessing the:
`echo Usage $0 file-to-save [project-name] ...`
Should be quoted too, so the `[`'s are not interpreted?

lluiscampos:
> I'm guessing the: `echo Usage $0 file-to-save [project-name] ...` Should be quoted too, so the `[`'s are not interpreted?

Thanks!!

PR #166, opened 2021-11-11
QA-235: Add new template .gitlab-ci-check-golang-lint.yml

lluiscampos:
This template is meant to replace .gitlab-ci-check-golang-static.yml,
but changing the linters there will terribly break all release branches
of many repositories.

Adding a new template instead and making a note on not using the old
one.

### DIFF
--- a/.gitlab-ci-check-golang-lint.yml
+++ b/.gitlab-ci-check-golang-lint.yml
@@ -1,0 +1,42 @@
+# .gitlab-ci-check-golang-lint.yml
+#
+# This gitlab-ci template performs all static checks from golangci-lint,
+# see https://github.com/golangci/golangci-lint
+#
+# Add it to the project in hand through Gitlab's include functionality
+#
+# include:
+#   - project: 'Northern.tech/Mender/mendertesting'
+#     file: '.gitlab-ci-check-golang-lint.yml'
+#
+# The standard .golangci.yml from this repository is used unless the
+# project including this template has its own version in the root dir.
+#
+# If the repo has some compile time dependencies, the template expect to
+# exist a deb-requirements.txt file with the Debian OS required packages
+#
+
+stages:
+  - test
+
+test:static:
+  stage: test
+  needs: []
+  except:
+    - /^saas-[a-zA-Z0-9.]+$/
+  image: golangci/golangci-lint:v1.43.0
+  before_script:
+    # Install compile dependencies
+    - if [ -f deb-requirements.txt ]; then
+        apt-get -qq update &&
+        apt install -yq $(cat deb-requirements.txt);
+      fi
+    # Copy .golangci.yml
+    - if [ ! -f .golangci.yml ]; then
+    -   curl -f
+          https://raw.githubusercontent.com/mendersoftware/mendertesting/master/.golangci.yml
+          --output .golangci.yml
+    -   sed -i "s/#CI_PROJECT_NAME#/${CI_PROJECT_NAME}/g" .golangci.yml
+    - fi
+  script:
+    - golangci-lint run -v
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,0 +1,47 @@
+run:
+  # timeout for analysis, e.g. 30s, 5m, default is 1m
+  timeout: 5m
+
+  # Skip linting _test.go files
+  tests: false
+
+  # Enables skipping of directories:
+  # vendor$, third_party$, testdata$, examples$, Godeps$, builtin$
+  skip-dirs-use-default: true
+
+  # Allow multiple parallel golangci-lint instances running.
+  # If false (default) - golangci-lint acquires file lock on start.
+  allow-parallel-runners: true
+
+linters:
+  enable:
+    - bodyclose
+    - deadcode
+    - errcheck
+    - gocyclo
+    - gofmt
+    - gosimple
+    - govet
+    - ineffassign
+    - lll
+    - staticcheck
+    - structcheck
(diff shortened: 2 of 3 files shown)


lluiscampos:
@alfrunes @oleorhagen what do you think? The task says to replace the contents, but I think that is a very problematic approach due to the release branches. I will follow-up updating master of all repos, but leave the release branches be.

Also, is this a good way to have an "standard" `.golangci.yml` but still let individual repos override it?

Thanks for your feedback.

.gitlab-ci-check-golang-lint.yml
```diff
@@ -0,0 +1,45 @@
...
+
+stages:
+  - test
+
+test:static:
+  stage: test
+  needs: []
+  except:
+    - /^saas-[a-zA-Z0-9.]+$/
+  image: golang:1.17
```
alfrunes, line 27:
```suggestion
  image: golangci/golangci-lint:v1.43.0
```

.gitlab-ci-check-golang-lint.yml
```diff
@@ -0,0 +1,45 @@
...
+  before_script:
+    # Install tool
+    - |
+      curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | \
+           sh -s -- -b $(go env GOPATH)/bin v1.31.0
+    # Install compile dependencies
+    - if [ -f deb-requirements.txt ]; then
+        apt-get -qq update &&
+        apt install -yq $(cat deb-requirements.txt);
+      fi
```
alfrunes, line 37:
```suggestion
```

.golangci.yml
```diff
@@ -0,0 +1,41 @@
...
+
+linters-settings:
+  gocyclo:
+    min-complexity: 20 # default is 30.
+
+  lll:
+    # max line length, lines longer will be reported. Default is 120.
+    line-length: 100
+    # tab width in spaces. Default to 1.
+    tab-width: 4
```
alfrunes, line 41: :sob:

alfrunes (changes requested):
Can we pin the golangci-lint version?

.golangci.yml
```diff
@@ -0,0 +1,41 @@
...
+    - staticcheck
+    - structcheck
+    - typecheck
+    - unused
+    - varcheck
+
+linters-settings:
+  gocyclo:
+    min-complexity: 20 # default is 30.
+
```
alfrunes, line 36: I'm not sure how to add this, but there's also a setting for `goimports` (see [1]) for sorting local imports (I'm sure @merlin would appreciate this).
https://github.com/mendersoftware/devicemonitor/blob/master/.golangci.yml#L38

.golangci.yml
lluiscampos: :ballot_box:

.gitlab-ci-check-golang-lint.yml
lluiscampos: :+1: Thanks

.golangci.yml
lluiscampos: I like it. Let me think a bit how can we use this in a generic way...

.gitlab-ci-check-golang-lint.yml
lluiscampos: Wait, the `deb-requirements` part needs to stay there for 3rd party dependencies that might be required for static analysis. For example [in deployments](https://github.com/mendersoftware/deployments/blob/master/deb-requirements.txt).

alfrunes: Do you need them for static checks as well? :thinking: 
In any case, I think the `golangci` image is built on top of a Debian image.

.golangci.yml
oleorhagen: ```-local=github.com/mendersoftware```
Would be nice at least (for `goimports`).

lluiscampos: @oleorhagen Yep :+1:

.gitlab-ci-check-golang-lint.yml
lluiscampos: Yes, we need them. I added them long time ago for it.

Okey, we will soon find out.

.golangci.yml
```diff
@@ -0,0 +1,47 @@
...
+linters-settings:
+  gocyclo:
+    # default is 30.
+    min-complexity: 20
+
+  goimports:
+    # to be edited by the template
+    local-prefixes:
+      "github.com/mendersoftware/#CI_PROJECT_NAME#"
```
lluiscampos, line 41: @oleorhagen This will pass to `goimoprts` the `local ...` flag

oleorhagen: Yes, but I would remove the `PROJECT_NAME`, and keep only the github/mendersoftware prefix, so that say `mender-artifact` is sorted alongside `mender` as our own project in say the mender-client.

But no strong feelings here ofc. It could just as well be counted as any other 3rd part import, I understand the idea.

oleorhagen (approved):

alfrunes (approved):

.golangci.yml
merlin-northern: @alfrunes I sure would! :)

.golangci.yml
mzedel: just since I've come across the related line-length setting... these settings here are directly against the motivating factors of the language > https://github.com/golang/go/issues/40028 < fmt and IMO this should be reversed...

PR #168, opened 2021-11-24
Do not check staging/hosted leaks unnecessary

oleorhagen:


check_commits.sh
```diff
@@ -133,14 +133,18 @@ function prevent_staging_and_hosted_leaks() {
     done
 }
 
+function branch_exists_in_remote() {
+    git remote show origin 2>/dev/null | grep -E "\b$1\b"
```
lluiscampos, line 137: Suggestion, make `grep` silent:
```suggestion
    git remote show origin 2>/dev/null | grep -q -E "\b$1\b"
```

lluiscampos (approved):
Looks good. Thanks for the fix

PR #169, opened 2021-11-25
QA-311 Check the contributor against the company map for all commit

oleorhagen:
Changelog: None

### DIFF
--- a/check_commits.sh
+++ b/check_commits.sh
@@ -28,6 +28,11 @@ do
             shift
             continue
             ;;
+        --contributor-email)
+            CHECK_CONTRIBUTOR_EMAIL=TRUE
+            shift
+            continue
+            ;;
         *)
             break
             ;;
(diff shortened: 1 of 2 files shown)


oleorhagen:
lol, least we know it's working

check_commits.sh
```diff
@@ -137,12 +142,36 @@ function branch_exists_in_remote() {
...
+    local -r user_email_company=$(echo ${user_email} | cut -d'@' -f2)
+    if [[ "${user_email_company}" =~ (northern.tech|cfengine.com) ]]; then
+        return 0
+    fi
+    company_map="$(wget --quiet --output-document=- https://raw.githubusercontent.com/mendersoftware/integration/master/extra/gitdm/company-map)"
+    if echo "${company_map}" | grep -q "${user_email_company}"; then
+        echo >&2 "Error: user email ${user_email} not found in the company map (integration/extra/gitdm/company-map)"
+        notvalid="$notvalid $commit"
+    fi
+}
```
lluiscampos, line 163: I am not sure about this. Some times we have people using personal e-mails (gmail, etc) and we don't add them to the `company-map` because we don't have a way to tell which company they are from.

:thinking: what was your idea for this @kacf ? We could make the check fail anyway (as this PR suggests) and then we force-merge this PRs knowing the reason for it.

lluiscampos (commented):
The code looks nice, by the way! I just added this comment to discuss a bit with Kristian before final review.

check_commits.sh
kacf: Yes, personal emails are quite common. I think this has to be a manual check before release, because both being in the company-map, and not being there, is valid. We could potentially solve that by having a "negative" list as well (`not-in-company-map.txt`), but personally I don't particularly like that approach.

Another disadvantage I can see with this check is that all new contributors will get a red build regardless of what they do, which is maybe not the best first impression?

oleorhagen: So, are we rejecting this ticket all together?

kacf: I would suggest so, unless anyone has a better idea.

lluiscampos: It is a shame we didn't discuss this in advance. That is mainly my fault and I am sorry for that.

At this point, I also think is best to reject :weary: 

Maybe [QA-310](https://tracker.mender.io/browse/QA-310) is enough to have an early alert of new contributors that need to be added to the company-map, so that the release checklist item is rather "check the tentative release info _here_ and potentially add missing contributor's companies".

oleorhagen: Alright, sounds good to me. Closing and rejecting.

kacf: > Maybe [QA-310](https://tracker.mender.io/browse/QA-310) is enough to have an early alert of new contributors that need to be added to the company-map, so that the release checklist item is rather "check the tentative release info _here_ and potentially add missing contributor's companies".

Maybe, but as a solution, I like [QA-342](https://tracker.mender.io/browse/QA-342) better, since it is tied to the pull request review, instead of being something on the side you need to check.

kacf: @lluiscampos: True, but maybe we should adapt it to submit both? My main point was that it's tied to the PR, it's not a separate rendering.

oleorhagen: I like the idea

lluiscampos: I added a comment in the task :+1:

PR #170, opened 2021-11-29
[RFC]: MEN-5241: Catch golang dependency modifications

oleorhagen:
This adds a check to catch unwanted changes to the monitor directory.

Changelog: None

### DIFF
--- a/.gitlab-ci-check-golang-lint.yml
+++ b/.gitlab-ci-check-golang-lint.yml
@@ -40,3 +40,18 @@ test:static:
     - fi
   script:
     - golangci-lint run -v
+
+test:govendor-check:
+  stage: test
+  needs: []
+  variables:
+    - GOLANG_VERSION=1.16
+  except:
+    - /^(master|hosted|staging|production|[0-9]+\.[0-9]+\.x|^saas-[a-zA-Z0-9.]+)$/
+  image: golang:${GOLANG_VERSION}
+  script:
+    - if [ -d vendor ]; then
+    -   go mod tidy
+    -   go mod vendor
+    -   git diff | wc -l && exit 1
+    - fi


.gitlab-ci-check-golang-lint.yml
```diff
@@ -40,3 +40,18 @@ test:static:
     - fi
   script:
     - golangci-lint run -v
+
+test:govendor-check:
+  stage: test
+  needs: []
+  variables:
+    - GOLANG_VERSION=1.16
```
oleorhagen, line 48: Keeping this open for multiple versions in different repos.

.gitlab-ci-check-golang-lint.yml
```diff
@@ -40,3 +40,18 @@ test:static:
...
+  except:
+    - /^(master|hosted|staging|production|[0-9]+\.[0-9]+\.x|^saas-[a-zA-Z0-9.]+)$/
```
oleorhagen, line 50: Not running on special branches

kacf (approved):
:+1:

.gitlab-ci-check-golang-lint.yml
lluiscampos: Okey, but shouldn't the default be today's latest?

.gitlab-ci-check-golang-lint.yml
```diff
@@ -40,3 +40,18 @@ test:static:
...
+  image: golang:${GOLANG_VERSION}
+  script:
+    - if [ -d vendor ]; then
+    -   go mod tidy
+    -   go mod vendor
+    -   git diff | wc -l && exit 1
```
lluiscampos, line 56: Okey. As an alternative, you could use [`git diff-index` instead](https://github.com/mendersoftware/mender-docs/blob/master/.gitlab-ci.yml#L77-L79).

.gitlab-ci-check-golang-lint.yml
```diff
@@ -40,3 +40,18 @@ test:static:
...
+test:govendor-check:
+  stage: test
+  needs: []
+  variables:
+    - GOLANG_VERSION=1.16
+  except:
+    - /^(master|hosted|staging|production|[0-9]+\.[0-9]+\.x|^saas-[a-zA-Z0-9.]+)$/
+  image: golang:${GOLANG_VERSION}
+  script:
+    - if [ -d vendor ]; then
```
lluiscampos, line 53: Are there any repos that have no `vendor` folder? That I can think of, only `integration-test-runner` has no vendor.

I think we can use the opportunity to align all, making this check expect a `vendor` folder and fail otherwise (if so, I think the `go mod vendor` + `git diff` will be sufficient, no need to explicitly check for the folder and fail)

lluiscampos (approved):
Looks good!

I commented on few things, but all of them are open for discussion or rejection.

.gitlab-ci-check-golang-lint.yml
oleorhagen: Perhaps, I just aligned with the already existing check here, which uses 1.16.
Another point is that we know different versions of the toolchain acts a little different, so should be aligned to the repo's version most likely (was my thining at least).
But I see your point with master, and thinking bout it, it's probably best :thinking:

.gitlab-ci-check-golang-lint.yml
oleorhagen: and `openssl` and `mender-cli` it seems, and to me aligning due to this seems a little unnecessary tbh, but then again, no hard feelings here.

lluiscampos: Okey, let's leave it. More flexible.

But `mender-cli` is a core component and I strongly recommend aligning it (can you create a JIRA task for it?).

.gitlab-ci-check-golang-lint.yml
lluiscampos: It's okey, you have a fair point too. Let's lock it to 1.16 for now.

.gitlab-ci-check-golang-lint.yml
oleorhagen: https://tracker.mender.io/browse/QA-357

oleorhagen:
Let's give it a spin!

mzedel (commented):
Since this relates to the product security we should not limit the checks to the latest `.gitlab-ci-check-golang-lint.yml`...
@oleorhagen please also add this to `.gitlab-ci-check-golang-static.yml` + to repos that are not aligned with the template (at least `mender` client, maybe others...)

lluiscampos:
@mzedel @oleorhagen,

The fix is now running, and it failed for example for mender-artifact, for which I launched a tentative PR:
* https://github.com/mendersoftware/mender-artifact/pull/369

>  @oleorhagen please also add this to `.gitlab-ci-check-golang-static.yml`

Given the experience above, this will fail for release branches. So unless we have a plan to follow-up in all repos, all release branches... I discourage to do this

> + to repos that are not aligned with the template (at least `mender` client, maybe others...)

This part yes, of course! But again, only for master IMO.

oleorhagen:
It's not running on release branches, remember the `except` clause :)

mzedel:
> @mzedel @oleorhagen,
> 
> The fix is now running, and it failed for example for mender-artifact, for which I launched a tentative PR:
> 
> * [Update dependencies: go mod tidy + go mod vendor mender-artifact#369](https://github.com/mendersoftware/mender-artifact/pull/369)
> 
> > @oleorhagen please also add this to `.gitlab-ci-check-golang-static.yml`
> 
> Given the experience above, this will fail for release branches. So unless we have a plan to follow-up in all repos, all release branches... I discourage to do this.
> 
> > * to repos that are not aligned with the template (at least `mender` client, maybe others...)
> 
> This part yes, of course! But again, only for master IMO.

let me repeat: this is a security issue and getting failures in the release branches is very much intentional, as we can't be sure if a past release contains a malicious change in one of the thousands of files committed. all of the previous releases that this check can be run against *should* be run against and if they fail, we have to either check the contents for malicious changes or report these as malicious releases.

lluiscampos:
> unless we have a plan to follow-up in all repos

So let's have a plan to follow-up in all repos and branches! What I think is very problematic is to "merge and forget" something like this.

mzedel:
the plan is not to merge and forget, but to merge and re-run all pipelines possible with this change and then be alerted by all failing builds to either issue warnings to our users or investigate and fix the dependencies.

oleorhagen:
Follow up: https://github.com/mendersoftware/mendertesting/pull/173

PR #173, opened 2021-11-30
MEN-5241: Follow ups

oleorhagen:
* Add the template to the static check as well
* Run the check on all branches, not dicounting released, hosted and staging branches.

### DIFF
--- a/.gitlab-ci-check-golang-lint.yml
+++ b/.gitlab-ci-check-golang-lint.yml
@@ -46,8 +46,6 @@ test:govendor-check:
   needs: []
   variables:
     GOLANG_VERSION: "1.16"
-  except:
-    - /^(master|hosted|staging|production|[0-9]+\.[0-9]+\.x|^saas-[a-zA-Z0-9.]+)$/
   image: golang:${GOLANG_VERSION}
   script:
     - if [ -d vendor ]; then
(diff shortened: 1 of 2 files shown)


lluiscampos (approved):
Thanks!

For release branches of `mender` and `mender-artifact`, we need to add `.gitlab-ci-check-golang-static.yml` to their pipelines [here](https://github.com/mendersoftware/mender/blob/3.1.x/.gitlab-ci.yml#L7) and [here](https://github.com/mendersoftware/mender-artifact/blob/3.6.x/.gitlab-ci.yml#L7).

.gitlab-ci-check-golang-lint.yml
```diff
@@ -46,8 +46,6 @@ test:govendor-check:
   needs: []
   variables:
     GOLANG_VERSION: "1.16"
-  except:
-    - /^(master|hosted|staging|production|[0-9]+\.[0-9]+\.x|^saas-[a-zA-Z0-9.]+)$/
```
mzedel, line 50: while possible for release branches, IMO no change should land in `hosted|staging|production|^saas-[a-zA-Z0-9.]+` that didn't go through the master pipeline first - so we could skip these runner minutes there

mzedel (approved):
Thank you!

.gitlab-ci-check-golang-lint.yml
oleorhagen: Better safe than sorry? I mean, the pipe is literally 10 seconds

mzedel: 🚀 in that case, sure - run it everywhere 🎉

PR #176, opened 2022-01-16
QA-363: Enforce conventional commits all throughout our code

oleorhagen:


check_commits.sh
```diff
@@ -10,12 +12,13 @@ do
             echo
             echo "    --signoffs    Enable checking of signoffs"
             echo "    --changelogs  Enable checking of changelogs"
+            echo "    --schema      Enable checking of the commit schema format (conventiona commits required)"
```
tranchitella, line 15:
```suggestion
            echo "    --schema      Enable checking of the commit schema format (conventional commits required)"
```

tranchitella (approved):

kacf (changes requested):
I have not looked at the code, but I don't think this should be merged yet. We have not had an introduction into how this works in practice, nor any discussion about how well this pattern works for our model. I'm very much in favor of the idea, and maybe it fits perfectly, but we need to have collectively done several things before this can be merged:

* Agreed that this will work for us, in detail, and applied any tweaks we see necessary.
* Made sure that everyone understands how to use it. This was the big mistake with changelogs, that people didn't understand how to use them for a long time (and I still see mistakes from time to time). This isn't all automatic, it needs correct input from people too. This could be a presentation, meeting or email, just some appropriate forum.

In addition to that, there are technical and user facing changes needed, such as our contribution guide, and possibly release-tool/changelog generator changes.

kacf:
> Agreed that this will work for us, in detail, and applied any tweaks we see necessary.

Just expanding on this a little bit: What I would like to see is a detailed proposal (which can include external docs if that is appropriate), with a list of the various keywords, how they need to appear in commit messages, and what effect they will have on our changelogs and release process.

oleorhagen:
> I have not looked at the code, but I don't think this should be merged yet. We have not had an introduction into how this works in practice, nor any discussion about how well this pattern works for our model. I'm very much in favor of the idea, and maybe it fits perfectly, but we need to have collectively done several things before this can be merged:
> 
>     * Agreed that this will work for us, in detail, and applied any tweaks we see necessary.

This we have already done, no?

I don't see that we need any tweaks, except that this will only cover our versioned repositories.

> 
>     * Made sure that everyone understands how to use it. This was the big mistake with changelogs, that people didn't understand how to use them for a long time (and I still see mistakes from time to time). This isn't all automatic, it needs correct input from people too. This could be a presentation, meeting or email, just some appropriate forum.
> 

To get this on the road, I will write up an email, with the announcement, and some resources.

> 
> In addition to that, there are technical and user facing changes needed, such as our contribution guide, and possibly release-tool/changelog generator changes.
>

Should be covered in here: https://tracker.mender.io/browse/QA-362

> 
> All these things need to be in place _before_ this PR can be merged.

Agreed.

kacf:
> This we have already done, no?

We started, but the proposal was more like "we should have RBAC!". Now we're talking about what shape it will take, and we still need to get multiple eyes on the details to make sure it fits us. No one got any introduction on the meeting to how it works.

Anyway, I saw you email, that's great, let's proceed from there!

oleorhagen:
> > This we have already done, no?
> 
> We started, but the proposal was more like "we should have RBAC!". Now we're talking about what shape it will take, and we still need to get multiple eyes on the details to make sure it fits us. No one got any introduction on the meeting to how it works.
> 

I did an introduction on friday, after stand-up to most of the team.

Also, I want to highlight that my thinking is along the lines of:

I don't think we can fully predict either how this will go, or what issues we have not predicted - which I think is okay, as this is orthogonal to the workflow we are doing at the moment, since we still have to do `changelogs`. So even if we start this workflow right now, it does not 'destroy' our current workflow; meaning the parts of the commits we ignore now, will simply be given meaning. If it doesn't work, then the changelogs are still there, and the rest of the commit is mute.

>
> Anyway, I saw your email, that's great, let's proceed from there!

Lets :)

oleorhagen:
Closed in favour of: https://github.com/mendersoftware/mendertesting/pull/179

PR #177, opened 2022-02-21
Documentation + check_license.sh fix

lluiscampos:


lluiscampos:
The latter is need for a `mender-artifact` dependabot update, [which fails with `license: Is a directory`](https://gitlab.com/Northern.tech/Mender/mender-artifact/-/jobs/2109449647).

oleorhagen (approved):

PR #181, opened 2022-03-30
chore: made binary extraction no longer rely on a shell

mzedel:
- this is to also support extracting files from scratch based images
Changelog: None

Signed-off-by: Manuel Zedel <manuel.zedel@northern.tech>


mzedel:
related to mendersoftware/inventory-enterprise#298, this should allow us to skip the check for `scratch` in the `Dockerfile`

lluiscampos (approved):

PR #182, opened 2022-04-25
fix(buildmaster): Making grammar link clickable

merlin-northern:
Changelog: none


oleorhagen (approved):
Thanks!

PR #184, opened 2022-04-27
fix: do not check commits on saas tags

tranchitella:
Changelog: none

Signed-off-by: Fabio Tranchitella <fabio.tranchitella@northern.tech>


alfrunes (approved):
ping @oleorhagen dependabot PRs are [blocking CI](https://gitlab.com/Northern.tech/Mender/create-artifact-worker/-/pipelines/526047923) on release tags

PR #186, opened 2022-04-28
Fix commitlint bug

oleorhagen:


commitlint/commitlint
```diff
@@ -237,6 +237,14 @@ function parse_footer() {
         if (match(tok, CHANGELOG)) {
             CHANGELOG_VAL=tok
             verify_changelog(tok)
+            # Consume the entire paragraph for the 'Changelog' if it is multiline
+            if (next_tok_type == "LINE") {
+                while (tok != EOF && next_tok_type == "LINE") {
+                    advance()
+                }
+                # Require empty line afterwards in paragraph
+                empty_line("A 'Changelog' over multiple lines needs to have a trailing empty line")
```
kacf, line 246: This won't barf if the changelog paragraph is the last line, will it?

Edit: Actually, I guess it will never happen, because signoff always comes after.

kacf (approved):
Btw, I saw in Slack that someone had a problem with letter casing also (`Changelog: title` instead of `Changelog: Title`). It would be nice if we could accept either of those. I'm just thinking of contributors, it saves an entire round trip if they make a mistake, and rejecting lowercase is kind of silly. Do you agree?

commitlint/commitlint
oleorhagen: Exactly :)

oleorhagen:
> Btw, I saw in Slack that someone had a problem with letter casing also (`Changelog: title` instead of `Changelog: Title`). It would be nice if we could accept either of those. I'm just thinking of contributors, it saves an entire round trip if they make a mistake, and rejecting lowercase is kind of silly. Do you agree?

Yeh, I agree. 

Let's fix :muscle:

oleorhagen:
@kacf https://github.com/mendersoftware/mendertesting/pull/187

PR #187, opened 2022-04-28
fix(commitlint): allow lowercase footer keyword values

oleorhagen:

### DIFF
--- a/commitlint/commitlint
+++ b/commitlint/commitlint
@@ -324,7 +324,7 @@ function lint_required_footer_values() {
 function verify_changelog(changelog) {
     # Strip the CHANGELOG[(fix|feat)] prefix
     gsub(CHANGELOG_PREFIX, "", changelog)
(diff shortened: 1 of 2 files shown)


commitlint/commitlint
```diff
@@ -324,7 +324,7 @@ function lint_required_footer_values() {
 function verify_changelog(changelog) {
     # Strip the CHANGELOG[(fix|feat)] prefix
     gsub(CHANGELOG_PREFIX, "", changelog)
-    if ( match(changelog, "(None|Title|Commit|All|\\S+(\\s+\\S+){2,})") == 0 ) {
+    if ( match(changelog, "([Nn]one|[Tt]itle|[Cc]ommit|[Aa]ll|\\S+(\\s+\\S+){2,})") == 0 ) {
```
kacf, line 327: This is ok, but I was thinking that we can just use case insensitive matchers, like `re.IGNORECASE` in Python. Not sure what the awk equivalent is. In general I don't think case is significant in any of these fields.

I think this would still not allow `changelog: title`, for example? And same for ticket. What do you think?

PR #199, opened 2022-06-15
chore: Show both stdout and stderr in shasum checks

MaciejTe:
chore: Show both stdout and stderr in shasum checks

Changelog: None
Ticket: None
Signed-off-by: Maciej Tomczuk <maciej.tomczuk@northern.tech>

### DIFF
--- a/check_license.sh
+++ b/check_license.sh
@@ -61,10 +61,8 @@ sed '/^$/d' $CHKSUM_FILE > $TMP_CHKSUM_FILE
 # Use the tmp-file for the rest of the script
 CHKSUM_FILE=$TMP_CHKSUM_FILE
 
-# Collect only stderr from the subcommand
 output="$(
-          exec 3>&1
-          shasum --warn --algorithm 256 --check $CHKSUM_FILE > /dev/null 2>&3
+          shasum --warn --algorithm 256 --check $CHKSUM_FILE
 )"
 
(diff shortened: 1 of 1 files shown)


oleorhagen:
Will probably be fairly verbose but

check_license.sh
```diff
@@ -61,10 +61,8 @@ sed '/^$/d' $CHKSUM_FILE > $TMP_CHKSUM_FILE
 # Use the tmp-file for the rest of the script
 CHKSUM_FILE=$TMP_CHKSUM_FILE
 
-# Collect only stderr from the subcommand
 output="$(
-          exec 3>&1
-          shasum --warn --algorithm 256 --check $CHKSUM_FILE > /dev/null 2>&3
+          shasum --warn --algorithm 256 --check $CHKSUM_FILE
```
kacf, line 65: Hmm, is this right? Previously stderr was captured by the `$()` because of the `exec 3>&1` and `2>&3` (not sure why the `exec` is needed though), but now it is not redirected anywhere and escapes the capture. This means that the `grep` below will probably never trigger.

oleorhagen: Good catch @kacf !

MaciejTe: Thanks @kacf , for now I'll convert this PR to draft so no one merges it

PR #206, opened 2022-08-15
fix(commitlint): no space allowed between : and title in the header

oleorhagen:
Changelog: None
Ticket: None

### DIFF
--- a/commitlint/commitlint
+++ b/commitlint/commitlint
@@ -32,6 +32,7 @@ BEGIN {
 
     ## special
     COLON = "^:"
+    SINGLE_SPACE = "^ {1}"
     EOF = "E_O_F"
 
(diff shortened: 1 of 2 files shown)


kacf (approved):
I have to be honest: I don't understand this at all! 🙂

Why did it allow arbitrary number of spaces (including zero) before? Is it part of the next token, `CHANGELOG`?

oleorhagen:
> I have to be honest: I don't understand this at all! slightly_smiling_face
> 
> Why did it allow arbitrary number of spaces (including zero) before? Is it part of the next token, `CHANGELOG`?

See here: https://github.com/mendersoftware/mendertesting/blob/commitlint/commitlint/commitlint#L190

The title parsing is/was pretty liberal :)

PR #208, opened 2022-08-18
fix: handle keywords in the title header

oleorhagen:


mzedel (approved):
such reliance on globals 🔥, but the tests pass, so ✅

PR #215, opened 2022-09-14
chore: Licence checker on source files adjustment for 1.19 go fmt

merlin-northern:
* in case a licence check on a source file fails, replace tab with 4 spaces and try again

Ticket: QA-441

### DIFF
--- a/check_license_source_files.sh
+++ b/check_license_source_files.sh
@@ -131,6 +131,8 @@ check_file() {
     local lines
     local lic_type
     local -r CM="$2"
+    local rc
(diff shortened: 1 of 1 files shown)


merlin-northern:
@alfrunes it is a bit naïve, but maybe it is enough? strange thing: when I noticed go fmt in 1.19 breaking the check it only happened for certain _test files, and still I cant produce it to re-format other files with tabs.

alfrunes (approved):

alfrunes (approved):

PR #217, opened 2022-09-23
fix(check_commits.sh): Use "range notation" for git rev-parse command

lluiscampos:
Instead of "symmetric difference notation".
    
The list of commits that we should prevent from leaking are all the
commits from staging/hosted branches that are not in the target branch
(aka range), but we should ignore other commits that are already in the
target branch (aka symmetric difference).
    
See: https://git-scm.com/docs/git-rev-parse#_dotted_range_notations


lluiscampos:
See [this pipeline](https://gitlab.com/Northern.tech/Mender/deviceauth/-/jobs/3074397733) for an example of the check misbehaving.

kacf (approved):

oleorhagen (approved):

oleorhagen (approved):

PR #223, opened 2022-11-10
fix: check licences: exclude certain files.

merlin-northern:
fix: check licences: exclude certain files.

license checker does not allow certain files to be created.
we explicitly exclude *.[c,go,cpp] from the check.

see:https://github.com/mendersoftware/devicemonitor/pull/167

Ticket: QA-492

### DIFF
--- a/check_license.sh
+++ b/check_license.sh
@@ -84,7 +84,7 @@ if echo "$output" | grep -q 'line is improperly formatted' -; then
 fi
 
 # Unlisted licenses not allowed.
-for file in $(find . -type f -iname 'LICEN[SC]E' -o -iname 'LICEN[SC]E.*' -o -iname 'COPYING'); do
+while read -r file; do
     file=$(echo $file | sed -e 's,./,,')
     # Files in ".COVERED_LICENSES" are omitted from checking. There are two main
     # reasons this is useful:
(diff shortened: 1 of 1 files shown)


lluiscampos:
@merlin-northern Please rename the commit title (and the PR title)

merlin-northern:
> @merlin-northern Please rename the commit title (and the PR title)

yes, I noticed. thank you. fixed.

check_license.sh
```diff
@@ -102,7 +102,7 @@ for file in $(find . -type f -iname 'LICEN[SC]E' -o -iname 'LICEN[SC]E.*' -o -in
         echo >&2 "$file has missing or wrong entry in $CHKSUM_FILE"
         ret=1
     fi
-done
+done < <(find . \( -type f -iname 'LICEN[SC]E' -o -iname 'LICEN[SC]E.*' -o -iname 'COPYING' \) -and -not -iname 'LICEN[SC]E.go' -and -not -iname 'LICEN[SC]E.c' -and -not -iname 'LICEN[SC]E.cpp')
```
kacf, line 105: Why put it down here, and use all this shell magic? It's harder to read...

kacf (commented):
Lol, that was unintended! Good catch.

check_license.sh
merlin-northern: iterating over find results is a common mistake.
for example:
```bash
#mkdir /tmp/d; for i in {1..4}; do touch /tmp/d/"ok $i"; done; for i in `find /tmp/d`; do echo "\"$i\""; done
"/tmp/d"
"/tmp/d/ok"
"1"
"/tmp/d/ok"
"2"
"/tmp/d/ok"
"3"
"/tmp/d/ok"
"4"
```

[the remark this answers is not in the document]
merlin-northern: sure, fixed.

merlin-northern:
last force push dedicated to Lluís -- another update to the commit message, and the github details.

check_license.sh
kacf: Ah, the `read` makes it line based. I didn't think of that. Clever!

kacf (approved):

check_license.sh
merlin-northern: thank you for the review!

PR #229, opened 2022-12-05
ci: Support customizing the filename path for image load and save

lluiscampos:
And revert #227


merlin-northern (approved):
what I had in mind is: use symlinks if possible so we can maybe maintain some level of back-compatibility. discussed privately.

PR #230, opened 2022-12-05
ci: Fix artifacts default path for build image

lluiscampos:
It seems like env variables in `artifacts` do not expand using bash syntax and the default path is evaluating to `/image.tar`.

Duplicate the paths instead, which will issue a warning on missing one of the two paths but that is acceptable.


olehermanse (approved):
Thanks for working on this!

PR #235, opened 2022-12-16
ci: Rename GitHub token to use for status updates

lluiscampos:


lluiscampos:
See also:
* https://github.com/mendersoftware/mender-qa/pull/649

lluiscampos:
@kacf The motivation is to be more explicit on the actual permissions that the token has. I also will be adding a `GITHUB_BOT_TOKEN_REPO_FULL` (protected) so it is good to have them clearly separated.

kacf (approved):
Good thinking! 👍

PR #239, opened 2022-12-20
fix: Use short arguments for sort and tail

lluiscampos:
It turns out that busybox implementation of sort and tail does not support the long arguments.

Amends commit 630089ee42519237d967093744134cbfbae6ab78


lluiscampos:
Example of a failure [here](https://gitlab.com/Northern.tech/Mender/mender-ci-workflows/-/jobs/3502105843).

oleorhagen (approved):

PR #240, opened 2023-01-04
Commitlint: Accept standard git revert commits

oleorhagen:
Also add better debug prints


commitlint/testcommitlint.sh
```diff
@@ -451,5 +451,11 @@ Changelog: none
 Signed-off-by: Kristian Amlie <kristian.amlie@northern.tech>"
 
 
+assert "true" \
+       "Accept standard git-revert commits" \
+"Revert build(docker-client): Add support for building the cpp-client
```
lluiscampos, line 456: In reality, the title of the old commit is quoted: `Revert "build(docker-client): Add support for building the cpp-client"`. Will this still work with the parser?

oleorhagen: hah, good catch! I copy pasted, and was like: Wtf is these extra quotes doing here...

oleorhagen: And it broke... :disappointed:

oleorhagen: Fixed.

One other thing, is that we probably shouldn't parse these commits at all.

lluiscampos: > One other thing, is that we probably shouldn't parse these commits at all.

We have to parse the commits. The `git revert` will only keep the title and "This reverts commit blablabla", but the user most likely appends some context on why the revert is there and possibly have a changelog.

lluiscampos (approved):

lluiscampos:
@oleorhagen Thank you for your contribution but we are very soon moving away from our custom commit checker :wink:

PR #245, opened 2023-02-07
QA-550: fix: Make sync:image job work when reusing the workspace

lluiscampos:
When reusing the workspace, a remote with name `github` already exists. Move instead to a temporary workspace.

See also related fix 4d17fc8.


oleorhagen (approved):
Nice!

Might not have to clone the whole tree though? :thinking:

lluiscampos:
> Might not have to clone the whole tree though? thinking

You are right, it should work with shallow clone. I'll amend, thanks!

lluiscampos:
@oleorhagen I found one issue, though. After a clone you need to `cd` into the new directory (before it was not needed)

oleorhagen:
Ohh.. Good point! :)

PR #253, opened 2023-05-04
chore: API docs lint fix.

merlin-northern:
api docs lint checks were not running due to:

WARNING: Entrypoint override disabled

and using the override itself.

* example of a job in master: https://gitlab.com/Northern.tech/Mender/deployments/-/jobs/4214764960
* see also, for running example of this patch: https://github.com/mendersoftware/deployments/pull/862

the problem with this fix: spectral lint returns 1 shall I add || true and if so, how to detect the failure?

Ticket: None

### DIFF
--- a/.gitlab-ci-check-apidocs.yml
+++ b/.gitlab-ci-check-apidocs.yml
@@ -51,11 +51,15 @@ test:validate-open-api:
     - mender-qa-worker-generic-light
   stage: test
   image:
-    name: stoplight/spectral
-    entrypoint: ['']
-  script: |
-    - echo "extends: ['spectral:oas']" > .spectral.yaml
-    - spectral lint -D -f junit -o spectral-report.xml docs/*.yml
+    name: alpine
+  before_script:
+    - apk --update add curl
+    - curl -L https://raw.github.com/stoplightio/spectral/master/scripts/install.sh -o install.sh
+    - sh install.sh
+  script:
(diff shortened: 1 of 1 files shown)


merlin-northern:
@kjaskiewiczz I ping you as the current BM.
@mzedel I am unsure what to do with the exit code of 1, I tested manually and the lint command from `spectral` seems to return 1 at least for deployments docs (I checked only internal docs, and management docs, all the same). so either it is just like that, or `spectral` is trying to tell us something. either way, with this patch I managed to get the linter to run, see here: https://github.com/mendersoftware/deployments/pull/862.

mzedel (approved):
Yeah the test fails because almost all our API specs have issues, for this reason this job is allowed to fail for now...
Thanks a lot for fixing the job definition, I guess I must have tried it locally only and didn't anticipate the spectral docs example to be faulty...

merlin-northern:
> Yeah the test fails because almost all our API specs have issues, for this reason this job is allowed to fail for now... Thanks a lot for fixing the job definition, I guess I must have tried it locally only and didn't anticipate the spectral docs example to be faulty...

thanks for the review and for the initiative to clean up the API specs!

mzedel:
> thanks for the review and for the initiative to clean up the API specs!

all just selfish prep for https://github.com/mendersoftware/gui/pull/3654 😄

PR #257, opened 2023-06-12
fix: fixed an issue that prevented swagger-cli from running through npx

mzedel:
Ticket: None
Changelog: None

### DIFF
--- a/.gitlab-ci-check-apidocs.yml
+++ b/.gitlab-ci-check-apidocs.yml
@@ -44,6 +44,7 @@ test:apidocs:verify-swagger:
   image: node:alpine
   script:
     # Verify that the Swagger docs are valid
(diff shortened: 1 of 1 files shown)


alfrunes (approved):

.gitlab-ci-check-apidocs.yml
```diff
@@ -44,6 +44,7 @@ test:apidocs:verify-swagger:
   image: node:alpine
   script:
     # Verify that the Swagger docs are valid
+    - npm i -g @apidevtools/swagger-cli
```
merlin-northern, line 47: that should be in the before script, right? (nit pick)

mzedel: this is what `npx` does internally, so it was always part of `script` 🤷‍♂️

PR #259, opened 2023-06-23
ci(go/units): Add TEST_MONGO_URL to unit tests job template

alfrunes:
Allows us to update debian image once mendersoftware/go-lib-micro#201 is merged and pulled to all backend services.


alfrunes (commented):
Side note: In the next iteration we need to create another version of this job to make this job work with older branches. Another approach would be to start versioning `mendertesting`.

mzedel (approved):
👍 also to the "we might need versioning for this"

PR #260, opened 2023-06-27
chore: Create a simplified v2 job template for Go unittests

alfrunes:


.gitlab-ci-check-golang-unittests-v2.yml
```diff
@@ -0,0 +1,113 @@
...
+    GIT_STRATEGY: clone # clone entire repo instead of reusing workspace
+    GIT_DEPTH: 0 # avoid shallow clone, this test requires full git history
+    TEST_MONGO_URL: "mongodb://mongo"
+  before_script:
+    # Install JUnit test reporting formatter
+    - go install github.com/jstemmer/go-junit-report@v1.0.0
+
+  script:
+    - >-
+      go test ./... -v -covermode=atomic -coverprofile=coverage.txt 2>&1 |
```
lluiscampos, line 57: Does this collect the coverage of all packages? Why then are we doing it [this complicated way](https://github.com/mendersoftware/mendertesting/blob/9ed85cd33d5ead852b3b7fd7b589b95debd854d6/.gitlab-ci-check-golang-unittests.yml#L86) in the old template?

.gitlab-ci-check-golang-unittests-v2.yml
```diff
@@ -0,0 +1,113 @@
...
+    # Coveralls env variables:
+    #  According to https://docs.coveralls.io/supported-ci-services
+    #  we should set CI_NAME, CI_BUILD_NUMBER, etc. But according
+    #  to goveralls source code (https://github.com/mattn/goveralls)
+    #  many of these are not supported. Set CI_BRANCH,
+    #  and pass few others as command line arguments.
+    #  See also https://docs.coveralls.io/api-reference
+    - export CI_BRANCH=${CI_COMMIT_BRANCH}
+  script:
+    - tar -xvf unit-coverage.tar
```
lluiscampos, line 91: This you don't need, the only artifact now is the coverage.txt

.gitlab-ci-check-golang-unittests-v2.yml
```diff
@@ -0,0 +1,113 @@
...
+#
+# By default, it installs MongoDB version 4.4 for Debian buster. You can
+# specify the Debian release name for the upstream APT repository and
+# the MongoDB version overriding the variables below:
+#
+# test:unit:
+#   variables:
+#     MONGODB_DEBIAN_RELEASE: "stretch"
+#     MONGODB_VERSION: "3.6"
+#
```
lluiscampos, line 31: Obsolete?

lluiscampos (approved):
Looking good! A bunch of nitpicks and questions :eyes: Pre-approving as I haven't spotted nothing really blocking.

.gitlab-ci-check-golang-unittests-v2.yml
alfrunes: Yes it does, but it omits packages that don't have `_test.go` files, which I believe is what we want anyway. I have no idea why we run go test for each individual package, to me it just looks like unnecessary boilerplate, but maybe it has something to do with how coveralls collects coverage?

.gitlab-ci-check-golang-unittests-v2.yml
alfrunes: Oops, forgot to adjust this.

.gitlab-ci-check-golang-unittests-v2.yml
alfrunes: yes

.gitlab-ci-check-golang-unittests-v2.yml
lluiscampos: The run for each package is older than coveralls so it cannot be coverage service specific... I also see it as unnecessary complex way of running the tests, but TBH never really questioned it :sweat_smile: 

I think it is safe to merge as-is: this is a new template so we will need to explicitly include it in the child repos with a PR, and then we could potentially see if there is anything missing (a drop in the coverage).

PR #261, opened 2023-06-27
ci: Move from `cp` to `rsync` for coverage artifacts collection

lluiscampos:
The `--parents` option that we intended to use is not available for Mac OS `cp` command. We can achieve the same functionality with `rsync`.

Inspired by:
* https://stackoverflow.com/questions/11246070/cp-parents-option-on-mac

### DIFF
--- a/.gitlab-ci-check-golang-unittests.yml
+++ b/.gitlab-ci-check-golang-unittests.yml
@@ -84,7 +84,7 @@ test:unit:
     - cd /go/src/github.com/mendersoftware/${CI_PROJECT_NAME}
   script:
     - go list ./... | grep -v vendor | xargs -n1 -I {} go test -v -covermode=atomic -coverprofile=../../../{}/coverage.txt {} 2>&1 | tee /dev/stderr | go-junit-report > ${CI_PROJECT_DIR}/test-results.xml || exit $?
-    - mkdir -p tests/unit-coverage && find . -name 'coverage.txt' -exec cp --parents {} ./tests/unit-coverage \;
+    - mkdir -p tests/unit-coverage && find . -name 'coverage.txt' -exec rsync -R {} ./tests/unit-coverage \;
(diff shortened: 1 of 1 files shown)


.gitlab-ci-check-golang-unittests.yml
```diff
@@ -84,7 +84,7 @@ test:unit:
     - cd /go/src/github.com/mendersoftware/${CI_PROJECT_NAME}
   script:
     - go list ./... | grep -v vendor | xargs -n1 -I {} go test -v -covermode=atomic -coverprofile=../../../{}/coverage.txt {} 2>&1 | tee /dev/stderr | go-junit-report > ${CI_PROJECT_DIR}/test-results.xml || exit $?
-    - mkdir -p tests/unit-coverage && find . -name 'coverage.txt' -exec cp --parents {} ./tests/unit-coverage \;
+    - mkdir -p tests/unit-coverage && find . -name 'coverage.txt' -exec rsync -R {} ./tests/unit-coverage \;
```
merlin-northern, line 87: as long as we are moving from cp, we can also drop find and use rsync `--include` and `--exclude` combination to transfer  all in one command.

lluiscampos: Will `rsync` on Mac OS support such options? And can you give me an explicit example of how to use it?

merlin-northern: something like:
```suggestion
    - mkdir -p tests/unit-coverage && rsync -avR --exclude=vendor --exclude=bin --exclude=tests --include="*/" --include=coverage.txt --exclude="*" . ./tests/unit-coverage/
```

mzedel: And for the support question: this seems to actually be supported on macOS 👌

lluiscampos:
Closing in favor of #260.

PR #265, opened 2023-06-28
Docker multiplatform build

oldgiova:
## Multiplatform build
* Required to Ticket: ALV-27 (Alvaldi-docs)
* Requires the base image build: https://github.com/mendersoftware/mender-test-containers/pull/362
* New `build:docker-multiplatform` build job added, it runs only with `MULTIPLATFORM_BUILD: "true"` 
* Can't OCI images with multi-arch manifests in the Gitlab Artifacts, so I'm using the Gitlab Registry with `--push` by pushing the image with the tag `${CI_PIPELINE_ID}` 
* The publish job is using `regclient` which has multi-arch manifests capabilities.
* With registry caching we can save some time:
```
      --cache-to type=registry,ref=${CI_REGISTRY_IMAGE}:ci_cache,mode=max
      --cache-from type=registry,ref=${CI_REGISTRY_IMAGE}:ci_cache
```

Also please notice that the source Dockerfile could be cross-compilation compliant, e.g.:
```
FROM --platform=$BUILDPLATFORM node:alpine AS build
ARG TARGETPLATFORM
```

### DIFF
--- a/.gitlab-ci-check-docker-build.yml
+++ b/.gitlab-ci-check-docker-build.yml
@@ -40,6 +40,23 @@
 #   * PUBLISH_IMAGE_DIGEST
 #   * PUBLISH_IMAGE_MENDER_DIGEST
 #
+#
+# Multiplafrom build:
+# Excludes build:docker and publish:image.
+#
+# Required variables:
+# MULTIPLATFORM_BUILD: "true"
+# MULTIPLATFORM_PLATFORMS: "linux/amd64,linux/armv7"
+#
+# Note: since jobs are evaluated before variables,
+# you have to override the build:docker job on the parent .gitlab-ci.yml;
+# otherwise you'll get both build:docker and build:docker-multiplatform jobs.
+# E.g.:
+#
+# build:docker:
+#   rules:
+#     - when: never
+
 
 stages:
   - build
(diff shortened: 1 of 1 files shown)


mzedel:
> * The publish job is using `regclient` which has multi-arch manifests capabilities.

🎉

> ```
> 
> Also please notice that the source Dockerfile has to be multiplaform compliant, e.g.:
> ```
> FROM --platform=$BUILDPLATFORM node:alpine AS build
> ARG TARGETPLATFORM
> ```

Should this not work automatically if you build with buildx? 🤔

alfrunes (approved):

oldgiova:
> > * The publish job is using `regclient` which has multi-arch manifests capabilities.
> 
> tada
> 
> > Also please notice that the source Dockerfile has to be multiplaform compliant, e.g.:
> > ```
> > FROM --platform=$BUILDPLATFORM node:alpine AS build
> > ARG TARGETPLATFORM
> > ```
> 
> Should this not work automatically if you build with buildx? thinking

Actually yes, let me correct my statement:
You might want to modify the Dockerfile to be cross-compilation compliant. 

Reference: https://docs.docker.com/build/building/multi-platform/

> Finally, depending on your project, the language that you use may have good support for cross-compilation. In that case, multi-stage builds in Dockerfiles can be effectively used to build binaries for the platform specified with --platform using the native architecture of the build node

mzedel (approved):

PR #278, opened 2023-08-28
ci: Allow to fail Mender specific jobs on build/publish template

lluiscampos:
This template is used by many repositories that build and publish Docker images. For Mender specific images, there are extra jobs that deal with Mender specific use cases.

Ideally these jobs should be disabled (removed from the pipeline) but so far we were only early exiting leaving the job "green".

With this change we return an specific error code to fail the job while we allow for the whole pipeline to continue.

The motivation is to better visualize what the job has done (nothing) instead of visualizing it as just success (while doing nothing).

### DIFF
--- a/.gitlab-ci-check-docker-build.yml
+++ b/.gitlab-ci-check-docker-build.yml
@@ -215,17 +215,19 @@ publish:image:mender:
     - *export_docker_vars
     - *get_release_tool_alpine
     - *docker_login_registries
+  allow_failure:
+    exit_codes: 137
   script:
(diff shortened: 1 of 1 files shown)


lluiscampos:
You can see an example of how it will look [here](https://gitlab.com/Northern.tech/Mender/mender-stress-test-client/-/pipelines/983810244).

mzedel (approved):

oldgiova (approved):

oldgiova:
Really nice!

PR #281, opened 2023-09-13
ci: Create dedicated Gitlab CI template for build/publish docker

alfrunes:


.gitlab-ci-docker-build-multiplatform.yml
```diff
@@ -0,0 +1,90 @@
...
+        --platform "$MULTIPLATFORM_TARGETS"
+    - |-
+      docker buildx build \
+        --cache-to "type=registry,ref=${CI_REGISTRY_IMAGE}:ci_cache,mode=max" \
+        --cache-from "type=registry,ref=${CI_REGISTRY_IMAGE}:ci_cache" \
+        --file "${DOCKER_DIR:-.}/${DOCKERFILE}" \
+        --build-arg GIT_COMMIT_TAG="${DOCKER_PUBLISH_COMMIT_TAG}" \
+        --platform "$MULTIPLATFORM_TARGETS" \
+        --provenance false \
+        --tag ${CI_REGISTRY_IMAGE}:${CI_PIPELINE_ID} \
```
oldgiova, line 67:
```suggestion
        --tag ${GITLAB_REGISTRY_TAG} \
```

.gitlab-ci-docker-build-multiplatform.yml
```diff
@@ -0,0 +1,90 @@
...
+  - build
+  - publish
+
+build:docker:
+  tags:
+    - mender-qa-worker-generic-light
+  stage: build
+  needs: []
+  variables:
+    DOCKER_BUILDKIT: 1
```
oldgiova, line 44: Some containers like the `workflows-worker` uses different tag but are starting from the same ${CI_REGISTRY_IMAGE}. By moving the tag definition to the variables section, we could easily override the tag for this kind of images.

```suggestion
    DOCKER_BUILDKIT: 1
    GITLAB_REGISTRY_TAG: '${CI_REGISTRY_IMAGE}:${CI_PIPELINE_ID}'
```

oldgiova (changes requested):
I'm insure about that. This seems not adding values to the [current feature](https://github.com/mendersoftware/mendertesting/blob/cd67132494005d35fd867bb4bb6d5540f2e3bc08/.gitlab-ci-check-docker-build.yml#L117).

I read that the main goal is to avoid masking the non-multiplatform jobs, by using this ad-hoc gitlab-ci. 
But the end goal of the multiplatfrom epic is to default to multiplatfrom and get rid of the previous [non-multiplatform](https://github.com/mendersoftware/mendertesting/blob/master/.gitlab-ci-check-docker-build.yml#L88) jobs.

So this is mainly a question of what future we want for the the current standard `.gitlab-ci-check-docker-build.yml`.

Please ask also some QA guy another opiinion.. @lluiscampos  maybe?

In any case, I'm requesting changes, since this won't work with the `workflows-worker` pipeline.

.gitlab-ci-docker-build-multiplatform.yml
alfrunes: Added `CI_IMAGE_REGISTRY_TAG_PREFIX` variable (not sure if built-in variables are available at that scope).

alfrunes:
> I read that the main goal is to avoid masking the non-multiplatform jobs, by using this ad-hoc gitlab-ci.
> But the end goal of the multiplatfrom epic is to default to multiplatfrom and get rid of the previous [non-multiplatform](https://github.com/mendersoftware/mendertesting/blob/master/.gitlab-ci-check-docker-build.yml#L88) jobs.

The easiest way of doing this is to create a new template and migrating existing repositories to the new template. By changing an old template this can cause a lot of trouble when backporting to dated branches if the template breaks backward compatibility. Because of the dependency on the release tool and different tagging strategy for nt-connect, I find the changes too intrusive.

.gitlab-ci-docker-build-multiplatform.yml
```diff
@@ -0,0 +1,93 @@
+variables:
```
lluiscampos, line 1: We are missing documentation like the rest of the templates.

.gitlab-ci-docker-build-multiplatform.yml
```diff
@@ -0,0 +1,93 @@
...
+  # Required variables
+  DOCKER_REGISTRY:
+    description: "The docker registry for publishing the image (e.g. 'docker.io')"
+  DOCKER_REPOSITORY:
+    description: "The docker registry to publish the image (e.g. 'northerntechhq/northern.tech')"
+  DOCKER_USERNAME:
+    description: "The username for DOCKER_REGISTRY"
+  DOCKER_PASSWORD:
+    description: "The password for DOCKER_REGISTRY"
+
```
lluiscampos, line 11: Thank you for adding the description field :+1:

.gitlab-ci-docker-build-multiplatform.yml
lluiscampos, line 1: Also the name of the file does not follow pattern of the rest of the templates.

lluiscampos (changes requested):
> > I read that the main goal is to avoid masking the non-multiplatform jobs, by using this ad-hoc gitlab-ci.
> > But the end goal of the multiplatfrom epic is to default to multiplatfrom and get rid of the previous [non-multiplatform](https://github.com/mendersoftware/mendertesting/blob/master/.gitlab-ci-check-docker-build.yml#L88) jobs.
> 
> The easiest way of doing this is to create a new template and migrating existing repositories to the new template. By changing an old template this can cause a lot of trouble when backporting to dated branches if the template breaks backward compatibility. Because of the dependency on the release tool and different tagging strategy for nt-connect, I find the changes too intrusive.

We have used both approaches in the past.

While I was Head of Mender QA, we discussed this issues in meetings and decided in case by case basis - sometimes it made sense to let pipelines fail and us to fix reactively (for example formatting checks) and other times it was better to create new templates (for example during linting tools upgrade).

In this specific case my personal opinion _as a developer_ is that it would have been better to make the multi platform build and publish it as a separate template (which could inherit from the existing one or be completely new). However this was done some sprints ago (with a [JIRA task](https://northerntech.atlassian.net/browse/QA-613)) and it was then that we should have discussed as a team what is the best approach.

Given today's state of things, I think is best to keep it as-is and not introduce this new template and migrate all the affected repos to it.

If you @alfrunes still believe that we need them separate and that this change was very intrusive, please discuss it in a Mender QA meeting and create a JIRA task to groom it.

.gitlab-ci-docker-build-multiplatform.yml
alfrunes: Thought there was no need since the variable section is pretty self documenting.

alfrunes:
... then what do I do with `nt-connect` (a non-mender client component)?

.gitlab-ci-docker-build-multiplatform.yml
lluiscampos: Sure. But then we should update the rest of the templates or, at a minimum, create a follow-up task for it.

lluiscampos:
> ... then what do I do with `nt-connect` (a non-mender client component)?

I think we discussed this already over Slack. But here my summary,

The template has two sets of jobs:
* `publish:image` and `publish:image-multiplatform` publish any image: mender repos, docs, cfengine things, alvaldi, etc
* `publish:image:mender` and `publish:image-multiplatform:mender` do an additional publishing with `-mender-1.3.4` kind of tag for Mender only repos. It uses the `release_tool` to figure out if the job is mender or not, and it will exit with a warning when it is not. --> For `nt-connect` you don't need these jobs, so you can either let them be with the warning or disable them with a  `rules: never`.

I created [QA-622](https://northerntech.atlassian.net/browse/QA-622) as a follow-up.

PR #282, opened 2023-09-13
Multiplatform prod rolling update process

oldgiova:
Ticket: QA-613

When a new production tag is issued:
* build and push the container image from the tag
* run the helm version bump for both staging and production.

* Why?
The job `trigger:alvaldi-helm-version-bump:staging` needs `publish:image-multiplatform` to finish successfully, while the original pipeline has just the `publish:image:saas` which is a re-tag of the `staging_commitsha` to `saas-v2023.....`

* Why I'm not simply creating creating a new `publish:image-multiplatform:saas` job which is doing the same but with `regctl`?
Because the helm version bump needs a dependency, to work properly in both master commits and production tags. This dependency could only be the `publish:image-multiplatform` job

### DIFF
--- a/.gitlab-ci-check-docker-build.yml
+++ b/.gitlab-ci-check-docker-build.yml
@@ -126,8 +126,6 @@ build:docker-multiplatform:
   rules:
     - if: $MULTIPLATFORM_BUILD != "true"
       when: never
-    - if: '$CI_COMMIT_TAG =~ /^saas-[a-zA-Z0-9.]+$/'
-      when: never
     - when: on_success
   image: "registry.gitlab.com/northern.tech/mender/mender-test-containers:docker-multiplatform-buildx-v1-master"
   services:
@@ -183,7 +181,7 @@ publish:image-multiplatform:
   rules:
     - if: $MULTIPLATFORM_BUILD != "true"
       when: never
-    - if: '$CI_COMMIT_BRANCH =~ /^(master|staging|production|feature-.+)$/'
+    - if: '$CI_COMMIT_BRANCH =~ /^(main|master|staging|production|feature-.+|saas-[a-zA-Z0-9.]+)$/'
   image: "registry.gitlab.com/northern.tech/mender/mender-test-containers:docker-multiplatform-buildx-v1-master"
   services:
     - docker:20.10.21-dind


lluiscampos:
> Ticket: QA-613
> 
> When a new production tag is issued:
> 
>     * build and push the container image from the tag
> 
>     * run the helm version bump for both staging and production.
> 
>     * Why?
>       The job `trigger:alvaldi-helm-version-bump:staging` needs `publish:image-multiplatform` to finish successfully, while the original pipeline has just the `publish:image:saas` which is a re-tag of the `staging_commitsha` to `saas-v2023.....`
> 
>     * Why I'm not simply creating creating a new `publish:image-multiplatform:saas` job which is doing the same but with `regctl`?
>       Because the helm version bump needs a dependency, to work properly in both master commits and production tags. This dependency could only be the `publish:image-multiplatform` job

All this could be in your commit message :)

.gitlab-ci-check-docker-build.yml
```diff
@@ -126,8 +126,6 @@ build:docker-multiplatform:
   rules:
     - if: $MULTIPLATFORM_BUILD != "true"
       when: never
-    - if: '$CI_COMMIT_TAG =~ /^saas-[a-zA-Z0-9.]+$/'
-      when: never
```
lluiscampos, line 130: I have to admit that I don't follow the reasoning for why we need this. I am not up to date with the "prod rolling update process".

The original intention for not rebuilding on saas Git tags is that we want to make sure that what we publish into the container registry is exactly (bit by bit exact) what we have tested in HM staging (iow the containers images that had an specific git tag on it).

Is this principle now broken with your change?

oldgiova: > I have to admit that I don't follow the reasoning for why we need this. I am not up to date with the "prod rolling update process".
> 
> The original intention for not rebuilding on saas Git tags is that we want to make sure that what we publish into the container registry is exactly (bit by bit exact) what we have tested in HM staging (iow the containers images that had an specific git tag on it).
> 
> Is this principle now broken with your change?

Yes, the principle is broken. Then I have two proposals:
1. move the HM staging tests into the tag publishing job as well
2. I'll try a workaround for the pipeline to keep the principle and in some way the tag pipeline just promotes the change.

lluiscampos: My intention is to bring up the historical background, not necessarily to block the change.

> 
> Yes, the principle is broken. Then I have two proposals:
> 
>     1. move the HM staging tests into the tag publishing job as well

It is more about manual testing in HM staging.

I am not the owner of the HM release process, if @alfrunes is okey with re-building the images I won't object further.


>     2. I'll try a workaround for the pipeline to keep the principle and in some way the tag pipeline just promotes the change.

This would be ideal.

oldgiova: > It is more about manual testing in HM staging.

The staging rolling release process is not changed: every commit to master/main leads to a new staging deployment. Here the process is change only with:
* now: 
  1. `staging` branch merge: build + staging deploy
  2. new release `tag`: docker retag from the previous `staging_sha` to the new `tag` (no tests) - the container image deployed branch is `staging`
* proposed new change:
  1. `staging` branch merge: tests + staging deploy
  2. new release `tag`: build + docker tag to the new `tag` - the container image deployed branch is `tag`

lluiscampos: Discussed f2f.

I don't have an strong opinion and I am okey with eventually doing this (re-build and test instead of only re-tag). At this point it is up to HM and SRE teams to agree :)

oldgiova: > Discussed f2f.
> 
> I don't have an strong opinion and I am okey with eventually doing this (re-build and test instead of only re-tag). At this point it is up to HM and SRE teams to agree :)

Thank you for the nice discussion! Since I found a workaround it makes no sense to force the flow for now. Let's adapt it in the future ;)

PR #289, opened 2023-09-29
Export the image as docker artifact for the QA pipeline

oldgiova:
Ticket: QA-613

### DIFF
--- a/.gitlab-ci-check-docker-build.yml
+++ b/.gitlab-ci-check-docker-build.yml
@@ -155,6 +155,11 @@ build:docker-multiplatform:
       --provenance false
       --push
       ${DOCKER_DIR:-.}
+    - docker save ${GITLAB_REGISTRY_TAG} > ${IMAGE_ARTIFACT_FILENAME:-${CI_PROJECT_DIR}/image.tar}
+  artifacts:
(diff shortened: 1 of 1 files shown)


lluiscampos (approved):

kjaskiewiczz (approved):

.gitlab-ci-check-docker-build.yml
```diff
@@ -155,6 +155,11 @@ build:docker-multiplatform:
       --provenance false
       --push
       ${DOCKER_DIR:-.}
+    - docker save ${GITLAB_REGISTRY_TAG} > ${IMAGE_ARTIFACT_FILENAME:-${CI_PROJECT_DIR}/image.tar}
```
alfrunes, line 158: I'm not sure if this works without the `--load` argument to the `buildx build` command. Also, using `docker save` it will only save the image for the architecture executing the command since the local docker image store does support multi-platform images.

oldgiova:
Sorry, I forgot this opened.. This was based on the wrong assumption that the repo was building the actual pr image for the QA pipeline, but the correct flow is that the qa pipeline is running an independent `docker build -t deployments:pr`

PR #291, opened 2023-10-02
MEN-6517 - chore: slightly relaxed requirement for copyright year checks to ease maintenance

mzedel:


check_license_source_files.sh
```diff
@@ -188,10 +188,9 @@ EOF
         cat "$license" >&2
         TEST_RESULT=1
     else
-        copyright_modified=$(echo "${CM} Copyright <copyright_year> Northern.tech AS" | sed "s/<copyright_year>/$modified_year/g")
-        copyright_file="$(head -n 1 "$file")"
-        if [ "$copyright_modified" != "$copyright_file" ]; then
-            echo >&2 "!!! FAILED license check on ${orig_file}; make sure copyright year matches last modified year of the file ($modified_year)"
+        copyright_year=$(grep -oP "${CM} Copyright \K\d{4} Northern.tech AS")
```
kacf, line 191: Nothing is grepped here, `$file` is missing.

Also, I don't think `\K` will work since it includes the suffix portion, IOW the company name.

check_license_source_files.sh
```diff
@@ -188,10 +188,9 @@ EOF
...
+        if [ $copyright_year -lt $added_year ]; then
```
kacf, line 192: Since the failure condition is based on a positive match, any failure in parsing will lead to numeric error, and for this to pass in error. This includes when the license is simply missing. One possible fix is:

```bash
if [ $copyright_year -ge $added_year ]; then
    # Success!
    :
else
    echo >&2 "!!! FAILED license check on ${orig_file}; make sure copyright year is at least the year the file was first added to git ($added_year)"
    TEST_RESULT=1
fi
```

kacf (changes requested):
Checking added-year is 👍, this will ease the new year!

check_license_source_files.sh
kacf: Oh sorry, I didn't mean you could just remove it. I can't immediately see any way to do this with one grep alone, so how about:

```bash
        copyright_year=$(grep -oP "${CM} Copyright \d{4} Northern.tech AS" "$file" | grep -oP "\d{4}")
```

kacf (approved):

check_license_source_files.sh
mzedel: ah grep seems slightly tricky to rely on 🙈 - Mac, "regular" grep and the alpine/ busybox grep all with their own surprises 😅... do you think `grep -oE` would work for you as well?

kacf: I think `-E` should work. I believe `-P` was only necessary for `\K`.

mzedel: 😞 unfortunately `-E` is not available in the `golang`/ `debian` based grep (but it is on mac + alpine)... so now I'm installing `grep` in the license checking stage instead and at least locally that works 🤞

PR #292, opened 2023-10-02
chore: adjusted versioned folder preventing automation

mzedel:
with the default template (Project URL, License URL, detected License), any update to a dependency requires an adjustment to the license file (making it harder than before), thus reducing it to just the project URL + detected license this will not be necessary and only license changes will need adjustments...
once this is merged I'll follow up with PRs to all the repos that have the csv file already

### DIFF
--- a/.gitlab-ci-check-license.yml
+++ b/.gitlab-ci-check-license.yml
@@ -79,7 +79,7 @@ test:check-license:golang:
     - go install github.com/google/go-licenses@latest
     - apk add --no-cache git
   script:
-    - go-licenses report . --ignore github.com/mendersoftware > licenses.csv
+    - go-licenses report . --ignore github.com/mendersoftware --template go_licenses_format.tpl > licenses.csv
     - git diff --exit-code
 
 test:check-license-source:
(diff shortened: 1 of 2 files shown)


.gitlab-ci-check-license.yml
```diff
@@ -79,7 +79,8 @@ test:check-license:golang:
     - go install github.com/google/go-licenses@latest
     - apk add --no-cache git
   script:
-    - go-licenses report . --ignore github.com/mendersoftware > licenses.csv
+    - echo "{{ range . }}{{printf "%s,%s\n" .Name .LicenseName}}{{ end }}" >> format.tpl
```
lluiscampos, line 82: I suggest we check in `format.tpl` in this repo instead of generating in on the fly.

This way a developer can copy it to a repository to manually generate the licenses report (similarly to [code style rules](https://github.com/mendersoftware/mendertesting/blob/master/.clang-format) or [linting rules](https://github.com/mendersoftware/mendertesting/blob/master/.golangci.yml))

lluiscampos (approved):

.gitlab-ci-check-license.yml
mzedel: yeah there I wasn't sure what direction to take, the checked in file is the proper way, but for me the copying from the pipeline and pasting it locally seemed more accessible - but it should now be checked in 👌

PR #294, opened 2023-10-03
chore: made vendor check get skipped if no vendor folder available

mzedel:
recently I've come across the vendor check job taking a while to start ([here](https://gitlab.com/Northern.tech/Mender/deployments-enterprise/-/jobs/5212756466)), so this should skip it if no `vendor` folder is present


lluiscampos (approved):
Thank you

PR #299, opened 2023-10-17
fix(commitlint): Fix the definitions of WORD and COMMIT_SCOPE

vpodzime:
A word is not a sequence of arbitrary **printable** characters. It's shouldn't contain dashes either, but in our case it simplifies things. For example, commit scope can then be a word in parantheses.

This fixes issues with commit messages containing scope and parantheses in the title, e.g. when mentioning SomeFunction().


oleorhagen (approved):

vpodzime (commented):
ACK

PR #300, opened 2023-11-01
ci: Bump image versions of images used for static checks

alfrunes:


oleorhagen (approved):
And Golang version?

alfrunes:
yes

PR #302, opened 2023-11-14
chore: Ignore github.com/northerntechhq in license checks

alfrunes:


kacf (commented):
I don't think we can do this on a generic basis, because we have forks within our namespace that carry somebody else's copyright, such as https://github.com/mendersoftware/openssl. Is there a specific problem, or just convenience?

alfrunes:
It was a lazy workaround for northerntechhq/nt-connect#31. Perhaps it would be better to put the ignore rule in a variable...

kacf:
Maybe, but is it a big a big hassle to maintain our own licenses in there too?

FWIW, in Yocto we need to generate a license file containing all licenses, including our own, and we use the `LIC_FILES_CHKSUM` file to do so. I assume we are going to do the same for `licenses.csv` files, and then we cannot have any exceptions.

kacf:
> Since we're moving away from using vendored dependencies, we're no longer relying on the LIC_FILES_CHKSUM file for managing licenses for the backend (there is a rule switch for these jobs checking for the existence of `license.csv` and `LIC_FILES_CHKSUM`).

Hmm. I think vendoring makes no difference. Isn't nt-connect a device component? And during building, it uses the components specified in `go.mod` regardless of whether they are vendored or not, right?

For that we need the full license list, including our own licenses. Unless you want to change the tooling for Yocto to somehow insert these licenses into the build recipe despite that they don't exist in the license file. But this sounds incredibly error prone to me, and I don't think I'd be in favor.

> I changed the implementation to ignore only the root module of the repository instead of using fixed prefixes

I'm still not completely clear on what is being gained here. Isn't it completely trivial to add our own licenses to the `licenses.csv` file? I think it's not a good idea to make exceptions here, especially not for prefixes that work as wildcards.

alfrunes:
Aha! Sorry, I didn't get it at first. The reason this was introduced in the first place was to have a completely automated way of maintaining the license checksum file to lower the burden of maintaining dependencies for the plethora of backend repository. However, since this is a device component and we're probably going to replace mender-connect at some point, I see that it might be a bad idea to remove the vendor folder in the first place. Closing...

PR #312, opened 2023-12-18
fix: set `GIT_DEPTH` to 0 to avoid unit test failures when checking licenses

tranchitella:
Changelog: None
Ticket: None

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -10,6 +10,9 @@ variables:
   UNVERSIONED_REPOSITORY: "true"
 
 test:unit:
+  variables:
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -8,6 +8,7 @@ stages:
 
 variables:
   UNVERSIONED_REPOSITORY: "true"
+  GIT_DEPTH: 0 # avoid shallow clone, this test requires full git history
```
lluiscampos, line 11: To prevent propagating it to undesired places, it should be set on the `test:unit: ` job

lluiscampos (approved):

PR #314, opened 2024-01-17
ci: Bump mongodb version to 6.0

alfrunes:


merlin-northern (approved):
do we have 6.0 in integration as well? I am approving only hoping that they are supporting my (new) CPU and dont require extensions that are no longer present. :>

tranchitella (approved):

kjaskiewiczz (approved):

alfrunes:
> do we have 6.0 in integration as well? I am approving only hoping that they are supporting my (new) CPU and dont require extensions that are no longer present. :>

PR is still open, but all tests passed: https://github.com/mendersoftware/integration/pull/2549.
Unfortunately, I don't think there's any way to get around the missing AVX instructions issue :crying_cat_face:
If you're feeling brave, you could try running MongoDB inside qemu.

PR #315, opened 2024-01-28
ci: Run acceptance tests with a candidate base image

merlin-northern:
Three friends:
* https://github.com/mendersoftware/mender-qa/pull/723
* https://github.com/mendersoftware/mendertesting/pull/315
* https://github.com/mendersoftware/mender-test-containers/pull/495

Ticket: QA-505

(cherry picked from commit 2b101b91da74c5434783e103df6e123a0a0deca8)

### DIFF
--- a/.gitlab-ci-check-docker-acceptance.yml
+++ b/.gitlab-ci-check-docker-acceptance.yml
@@ -98,6 +98,11 @@ test:acceptance_tests:
     - if [ -n "$REGISTRY_MENDER_IO_USERNAME" ]; then
     -   docker login -u "$REGISTRY_MENDER_IO_USERNAME" -p "$REGISTRY_MENDER_IO_PASSWORD" registry.mender.io
     - fi
+    - if [ -n "$MENDER_TEST_CONTAINERS_CANDIDATE_TAG" ]; then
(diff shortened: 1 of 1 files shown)


oleorhagen (approved):
Why pass the pwd through stdin? :thinking:

lluiscampos (approved):
Please remove the `PoC` of the commit. Otherwise :+1:

lluiscampos:
> Why pass the pwd through stdin? 🤔

@oleorhagen `docker login` considers it insecure:
```
WARNING! Using --password via the CLI is insecure. Use --password-stdin.
```

IMO for CI does not really matter, as you have the password in the env in any case so the risk of leaking it is the same... but it is nice not to see the warning in the output ;)

oleorhagen:
> > Why pass the pwd through stdin? 🤔
> 
> @oleorhagen `docker login` considers it insecure:
> 
> ```
> WARNING! Using --password via the CLI is insecure. Use --password-stdin.
> ```
> 
> IMO for CI does not really matter, as you have the password in the env in any case so the risk of leaking it is the same... but it is nice not to see the warning in the output ;)

haha, yes, that's what I was thinking too!

And I agree, fewer warnings, less disruptions :)

PR #320, opened 2024-04-03
ci: Set empty `dependencies` to the status jobs

lluiscampos:
This is a nice to have optimization. See comment.

### DIFF
--- a/.gitlab-ci-github-status-updates.yml
+++ b/.gitlab-ci-github-status-updates.yml
@@ -38,12 +38,16 @@ stages:
 
 github:start:
   extends: .github_status_template
+  dependencies: []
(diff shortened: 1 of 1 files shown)


lluiscampos:
The motivation is to avoid post jobs like [this one](https://gitlab.com/Northern.tech/Mender/mender-test-containers/-/jobs/6528128967) that take several minutes to download Artifacts for no good use. I verified that it works properly [in here](https://github.com/mendersoftware/mender-binary-delta/pull/283).

kacf (approved):
Nice, I didn't know that `dependencies` could circumvent the problem with `needs`!

oldgiova (approved):

PR #329, opened 2024-06-04
ci: avoiding docker hub limits

oldgiova:
Logging into the docker hub in the build steps to avoid the docker hub limits

Ticket: None
Changelog: None


.gitlab-ci-check-docker-build.yml
```diff
@@ -142,6 +142,7 @@ build:docker-multiplatform:
     - docker:20.10.21-dind
   before_script:
     - docker login -u $CI_REGISTRY_USER -p $CI_REGISTRY_PASSWORD $CI_REGISTRY
+    - *docker_login_registries
```
lluiscampos, line 145: This won't work on non-protected branches, right? (it shouldn't, otherwise we can leak secrets...)

oldgiova: Right! Now I remember why I didn't implemented this before!

PR #331, opened 2024-07-12
chore: Publishing `saas-v*` tags updates the latest tag

alfrunes:


kjaskiewiczz (approved):

alfrunes:
I'm not sure what to do with open-source repos. I wanted to tag the latest semver version to `latest`, but found that these images are published from https://github.com/mendersoftware/mender-qa so I found it hard to constrain the publish to only the latest tag.

lluiscampos (changes requested):
This change we need to discuss more widely.

Up until now "latest" pointed always to the latest semantic version. I am fine with changing this (that is the whole point of the rolling release!).

However I think we should manage it only from the mono repo pipeline and not mess with these templates that are used wildly (even out of Mender product).

(I am requesting changes in the spirit of a discussion, not that I have an specific thing to change)

PR #335, opened 2024-09-02
Avoid docker login hangs when missing creds  

oldgiova:
Ticket: QA-752

Recently, a docker login failure is waiting for 15 min to wait for web browser login. This slows down a lot the pipelines that don't need login to docker, but they are reusing the mendertesting shared pipeline configurations.
With this change, we'll ignore the docker login if the variable is not set.

### DIFF
--- a/.gitlab-ci-check-docker-build.yml
+++ b/.gitlab-ci-check-docker-build.yml
@@ -78,10 +78,18 @@ stages:
   - publish
 
 .docker_login_registries: &docker_login_registries |
-  docker login -u "$REGISTRY_MENDER_IO_USERNAME" -p "$REGISTRY_MENDER_IO_PASSWORD" registry.mender.io || \
-    echo "Warning: registry.mender.io credentials unavailable or invalid"
-  docker login -u "$DOCKER_HUB_USERNAME" -p "$DOCKER_HUB_PASSWORD" || \
-    echo "Warning: Docker credentials unavailable or invalid"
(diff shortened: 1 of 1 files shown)


.gitlab-ci-check-docker-build.yml
```diff
@@ -78,10 +78,18 @@ stages:
   - publish
 
 .docker_login_registries: &docker_login_registries |
-  docker login -u "$REGISTRY_MENDER_IO_USERNAME" -p "$REGISTRY_MENDER_IO_PASSWORD" registry.mender.io || \
-    echo "Warning: registry.mender.io credentials unavailable or invalid"
-  docker login -u "$DOCKER_HUB_USERNAME" -p "$DOCKER_HUB_PASSWORD" || \
-    echo "Warning: Docker credentials unavailable or invalid"
+  if [ -n "$REGISTRY_MENDER_IO_USERNAME" ]; then \
```
lluiscampos, line 81: I think you should check for the PASSWORD variables instead, which is the variable that for sure will be protected and potentially not available.

lluiscampos (approved):
It looks good! I think we could change the if condition for less false positives but in any case looks safe this way :+1:

.gitlab-ci-check-docker-build.yml
oldgiova: Nice trick!

lluiscampos (approved):

oldgiova:
Force-pushed a newline in the commit comment, just to trigger the gitlab pipeline

PR #336, opened 2024-09-09
chore: Install setuptools when running apidocs check

alfrunes:


mzedel (approved):

mzedel:
Slightly longer term I would vote to default to `test:validate-open-api` by not allowing it to fail + removing the swagger + yaml jobs.

alfrunes:
Thanks for the swift review! :dash:

alfrunes:
> Slightly longer term I would vote to default to `test:validate-open-api` by not allowing it to fail + removing the swagger + yaml jobs.

Yup.. Would probably be a good idea to refactor the backend API specs into a single OpenAPI 3 doc and run a standardized tool and not a self-composed unmaintained python script.

PR #339, opened 2024-09-27
ci: monorepo support for helm version bump

oldgiova:
The Helm Version Bump is accepting multiple keys to update the tags.

Ticket: QA-760

* Add support for monorepo (from single `CONTAINER`  to multiple `CONTAINERS` input)
* Add DRY_RUN feature
* not changing the `repository` key: the one in the `values-[staging|prod].yaml` will apply

### DIFF
--- a/.gitlab-ci-check-helm-version-bump.yml
+++ b/.gitlab-ci-check-helm-version-bump.yml
@@ -3,22 +3,23 @@
 # This gitlab-ci template updates the Helm Chart versions
 # for a given container
 #
-# Requires DOCKER_REPOSITORY variable to be set in the calling Pipeline.
+# Requires DOCKER_REGISTRY_ADDRESS variable to be set in the calling Pipeline.
 # Add it to the project in hand through Gitlab's include functionality
 # variables:
-#   DOCKER_REPOSITORY: <Image FQN, i.e mendersoftware/reponame>
+#   DOCKER_REGISTRY_ADDRESS: registry.mender.io
 # include:
 #   - project: 'Northern.tech/Mender/mendertesting'
 #     file: '.gitlab-ci-check-helm-version-bump.yml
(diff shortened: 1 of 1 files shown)


alfrunes (commented):
Can we please move this to the monorepo?
By proceeding to use mendertesting as a template repository we will continue to break maintenance branches.

mzedel:
> Can we please move this to the monorepo?
> By proceeding to use mendertesting as a template repository we will continue to break maintenance branches.

The monorepo doesn't exist in isolation though and e.g. the bump is getting used in other repos. Should we instead maybe version the templates and treat them as regular dependencies to limit the breakage?

oldgiova:
> Can we please move this to the monorepo?
> By proceeding to use mendertesting as a template repository we will continue to break maintenance branches.

The helm version bump CI code is used by `alvaldi-helm` and `mender-helm` only. By moving it to monorepo, we should move also `mender-helm` to monorepo (something that we could do, actually, but we chosen not to do at the release-split start), and we need to include it from `alvaldi-helm` 

> 
> The monorepo doesn't exist in isolation though and e.g. the bump is getting used in other repos. Should we instead maybe version the templates and treat them as regular dependencies to limit the breakage?

I like the idea of versioning it. You mean to create a new tag for the `mendertesting` repo?, for example:
```
  - project: 'Northern.tech/Mender/mendertesting'
    file: '.gitlab-ci-check-helm-version-bump.yml'
    ref: v1.0.1
```

Or you mean to just rename the job to `helm-version-bump_v1:`, `helm-version-bump_v2:` and so on?

alfrunes:
Sorry, I was confused while reading the PR. I mistakenly thought these were templates imported by the mender-server project and was just a little paranoid to not repeat the same mistake we did before.

PR #343, opened 2024-10-24
chore: reduced duplicate commit body render directives 

mzedel:
to ensure commit context is present for scoped commits too


oldgiova (approved):
Nice! Thanks for cleaning up my mess!

oldgiova (approved):

PR #350, opened 2024-11-15
ci: Add quiet flag to upload/download S3 artifacts

lluiscampos:
This is noisy and can flood the CI logs.


lluiscampos:
An example of a flooded CI log can be found [here](https://gitlab.com/Northern.tech/Mender/mender-convert/-/jobs/8378072588).

oldgiova (approved):

PR #352, opened 2024-12-02
chore: don't redirect shasum command to null in check_license.sh 

danielskinstad:
This lets us see which files have the wrongly computed checksum, it prints:
* `shasum: WARNING: 1 computed checksum did NOT match` 
   `path/to/license/LICENSE: FAILED`

instead of just :
* `shasum: WARNING: 1 computed checksum did NOT match`


lluiscampos (approved):
I had this issue before and didn't investigate if they were flags for this... thanks for fixing :+1:

PR #355, opened 2024-12-23
ci: Remove skip-tags from cliff.toml

alfrunes:
Tags are managed by release-please so it will have to follow whatever is used there.


oldgiova (approved):

oldgiova:
Sorry, I remember now why this was added: because with this ignore rc tags, the new stable release will include all the changes from the beginning. Without this rc tag, new stable release will include only incremental changes before the last rc tag.
So I think this is not we want..

alfrunes:
This seems like a config that would be useful in a single use-case. Anyway, I'm closing this one and using a local override to allow release candidates.

PR #358, opened 2025-01-06
chore: bump golang version

danielskinstad:
We need a newer golang version in the unit tests, because go now allows patch-versions for the go version in the go.mod - which some newer dependencies require.


lluiscampos (approved):
Note that there is [a v2 version of this template](https://github.com/mendersoftware/mendertesting/blob/master/.gitlab-ci-check-golang-unittests-v2.yml), it would be best to update both.

PR #359, opened 2025-01-07
chore: bump mongodb in golang unit tests

danielskinstad:
We can't get libssl1.1 from the upstream Debian after updating the image.


alfrunes (approved):

.gitlab-ci-check-golang-unittests.yml
alfrunes, line 1: I would prefer using `.gitlab-ci-check-golang-unittests-v2.yml` and avoid installing mongodb alltogether.

PR #360, opened 2025-02-13
chore: extended .gitlab-ci-check-docker-build.yml build args

aleksandrychev:


oldgiova (approved):
Not sure if every `build:docker` jobs are requiring the GIT_COMMIT_BRANCH arg, but since this will not broke build jobs without that arg, it's ok for me

oldgiova (approved):

PR #362, opened 2025-03-24
Add tickets from ME repository to be visible in the changelog.

pasinskim:
This should help with further automating releases with Jira `fixedVesrsions`. It won't help external users as they won't be able to access the tickets but it is the same for other projects which are there already (QA, SEC).


oldgiova (approved):

pasinskim:
> 👌 - you might want to also add it here: https://github.com/mendersoftware/mendertesting/blob/master/utils/cliff.toml.scoped#L113-L117 and we might have to find a way to combine these later on I guess 😬

Thanks @mzedel. I was not aware of this one. Fixed!

PR #366, opened 2025-04-11
fix: Do not override EXTRA_DOCKER_ARGS from parent scope

lluiscampos:
Depending on the trigger of the build, the variable value from the template will get precedence, clearing the intended value of the user.

This is the case for example when running a manual pipeline through GUI, where this variable is blanked out. See a failure pipeline here:
* https://gitlab.com/Northern.tech/Mender/mender-convert/-/jobs/9696025686

### DIFF
--- a/.gitlab-ci-check-docker-build.yml
+++ b/.gitlab-ci-check-docker-build.yml
@@ -71,9 +71,6 @@ variables:
   MULTIPLATFORM_PLATFORMS:
     description: "Comma-separated list of targets to build os/arch[/<version>]"
     value: "linux/amd64,linux/arm64"
-  EXTRA_DOCKER_ARGS:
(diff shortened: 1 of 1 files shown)


oldgiova (approved):
Maybe we should add the `EXTRA_DOCKER_ARGS` variable to the Group Variables? To avoid breakages in case of some job relying on this as undefined.... Even though I don't think there are actual use cases..

lluiscampos:
> Maybe we should add the `EXTRA_DOCKER_ARGS` variable to the Group Variables? To avoid breakages in case of some job relying on this as undefined.... Even though I don't think there are actual use cases..

I believe that it won't break anything. Note for example `DOCKER_BUILD_ARGS` which is not defined and used in the legacy build job.

PR #367, opened 2025-05-15
QA-1032: Add CI/CD component commit-lint

lluiscampos:
The template uses `commitlint` to lint every commit that is unique to the current branch. The Git logic is a bit cumbersome, but due to our GitHub<->GitLab sync we don't have info on the base branch for the branch being tested.

### DIFF
--- a/commitlint.config.js
+++ b/commitlint.config.js
@@ -1,0 +1,21 @@
+module.exports = {
+  extends: ['@commitlint/config-conventional'],
+
+  rules: {
+    'body-max-line-length': [1, 'always', 100],
+    'subject-case': [1, 'always', ['lower-case', 'sentence-case']],
+    'signed-off-by': [2, 'always', 'Signed-off-by'],
+    'body-leading-blank': [2, 'always'], // body must be preceded by a blank line
+                                         // exit with error if not
+  },
+
+  helpUrl: `
+  Commit messages must follow conventional commit format:
+  https://www.conventionalcommits.org/en/v1.0.0/#summary
+      type(optional-scope): subject
+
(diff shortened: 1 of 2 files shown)


lluiscampos:
@oldgiova Thank you for your comments.

> That's really nice and looks we're progressing nicely with a module oriented approach, also in the pipeline.. Great job!
> 
> Just a minor observation: the Gitlab component could be versioned? I'm figuring out if a project needs a fix on this component, that breaks the other projects..

Yes! I shared in today's demo my idea that once we design the first version for the "3C release", we would then tag the repo with v1.0.0 for example. Then as we do updates or modifications we can keep versioning further without worrying about breaking maintenance branches.

> 
> Also, playing devil's advocate here: what's the additional value of a component versus the usual
> 
> ```
> include:
>   - project: 'Northern.tech/Mender/mendertesting'
> ```
> 
> ? I got that it could have much value, only when published to the Public Catalog, do we want to do this (I'm not against this, just double checking)

I shared also the key points during the demo. The main advantage is variable scoping, as with the spec/input design we isolate don't need to share variables and risk overriding what we don't want. See [this recent issue](https://github.com/mendersoftware/mendertesting/commit/a04b836de8891eb6540bd5fe6ed2c37922aecba4) as an example.

danielskinstad (approved):
Nice, looks good to me

oldgiova (approved):

mzedel (approved):
As we've spoken about in person I'm not a fan of the approach to get the commit diff, but since it's working for us for a while already => 🚀 and we can look into this on a later day 🙂

PR #375, opened 2025-07-02
fix: Ignore `brew remove` failing in brew build cleanup

vpodzime:
We want to run all the steps in `after_script` and `brew remove WHATEVER` can easily fail if the job actually failed to build or install the thing.

Ticket: MEN-8331
Changelog: none


danielskinstad (approved):
Makes sense

PR #378, opened 2025-07-22
feat: component to add workstation-tools-repository

danielskinstad:
Originally i wanted it to be a template which you extended, but that didn't work due to possible override of `before_script`. I instead made it an anchor, but to use that in files that include the component you can't do the regular `*add_workstation_tools_repository` so you need to use `!reference` https://docs.gitlab.com/ci/yaml/yaml_optimization/#reference-tags 

To use this you must include the component and then you could either do:
```
- !reference [.install_mender_artifact]`
```
or 
```  
- !reference [.add_workstation_tools_repository]
- apt-get install -y mender-artifact
```

### DIFF
--- a/templates/workstation-tools-repository.yml
+++ b/templates/workstation-tools-repository.yml
@@ -1,0 +1,42 @@
+spec:
+  inputs:
+    repo_channel:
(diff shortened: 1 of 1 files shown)


vpodzime (approved):
Really nice! :heart:

PR #392, opened 2025-09-12
ci: Add common module for golang licenses check

elkoniu:
Ticket: QA-1163

### DIFF
--- a/templates/release-oslicenses-golang.yml
+++ b/templates/release-oslicenses-golang.yml
@@ -1,0 +1,106 @@
+spec:
+  inputs:
+    remote_license_file:
+      description: Remote license file to be used (e.g., "302.Release-information/03.Open-source-licenses/30.mender-artifact/docs.md")
+    golang_version:
+      description: Golang version image to be used
+      default: 1.24
+    golang_flags:
+      description: Golang flags to be used
+      default: ""
+    stage:
+      description: The pipeline stage where to perform the license release
+      default: publish
+    runner:
+      description: GitLab CI runner tag
+      default: hetzner-amd-beefy
+    github_user_name:
+      description: GitHub user name
+      default: mender-test-bot
+    github_user_email:
+      description: GitHub user e-mail address
+      default: mender@northern.tech
+---
+
+release:golanglicenses:generate:
+  stage: $[[ inputs.stage ]]
+  tags:
+    - $[[ inputs.runner ]]
+  rules:
+    - changes:
+        paths: ["**/*"]
+        compare_to: $DEFAULT_BRANCH
+    - if: '$CI_COMMIT_REF_PROTECTED == "true"'
+      when: on_success
+  image: golang:$[[ inputs.golang_version ]]
+  needs: []
+  variables:
+    GOFLAGS: $[[ inputs.golang_flags ]]
+  before_script:
+    - go install github.com/google/go-licenses@v1.6.0
+    - if [ ! -f go-licenses.gotpl ]; then
+    -   curl --fail --output go-licenses.gotpl
+          https://raw.githubusercontent.com/mendersoftware/mendertesting/refs/heads/master/go-licenses.gotpl
+    - fi
+
+  script:
+    - go-licenses check --disallowed_types=forbidden,restricted,unknown --ignore=github.com/mendersoftware/$CI_PROJECT_NAME ./...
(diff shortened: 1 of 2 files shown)


templates/release-oslicenses-golang.yml
```diff
@@ -0,0 +1,106 @@
...
+release:golanglicenses:generate:
+  stage: $[[ inputs.stage ]]
+  tags:
+    - $[[ inputs.runner ]]
+  rules:
+    - changes:
+        paths: ["**/*"]
+        compare_to: $DEFAULT_BRANCH
+    - if: '$CI_COMMIT_REF_PROTECTED == "true"'
+      when: on_success
```
danielskinstad, line 34: I guess it doesn't hurt, but AFAICT this is the default value and therefore we don't need to specify 'when: on_success,. Am I missing something?

templates/release-oslicenses-golang.yml
```diff
@@ -0,0 +1,106 @@
...
+  image: golang:$[[ inputs.golang_version ]]
+  needs: []
+  variables:
+    GOFLAGS: $[[ inputs.golang_flags ]]
+  before_script:
+    - go install github.com/google/go-licenses@v1.6.0
```
danielskinstad, line 40: Perhaps we could extract the version to a variable in the spec?

templates/release-oslicenses-golang.yml
```diff
@@ -0,0 +1,106 @@
...
+    - if [ ! -f go-licenses.gotpl ]; then
+    -   curl --fail --output go-licenses.gotpl
+          https://raw.githubusercontent.com/mendersoftware/mendertesting/refs/heads/add-golang-license-template/go-licenses.gotpl
+    - fi
+
+  script:
+    - go-licenses check --disallowed_types=forbidden,restricted,unknown --ignore=github.com/mendersoftware/$CI_PROJECT_NAME ./...
+    - go-licenses report --template=./go-licenses.gotpl --ignore=github.com/mendersoftware/$CI_PROJECT_NAME ./... > licenses.md
+  artifacts:
+    when: on_success
```
danielskinstad, line 50: I guess the same applies here: what is the purpose of specifying on_success?

templates/release-oslicenses-golang.yml
```diff
@@ -0,0 +1,106 @@
...
+        echo "ERROR: Remote license file $[[ inputs.remote_license_file ]] must exist"
+        exit 1
+      fi
+    - git checkout -b licenses-${CI_JOB_ID}
+    - |
+      if [ ! -e ../.licenses_header.md ]; then
+        echo "ERROR: ../.licenses_header.md must exist"
+        exit 1
+      fi
+    - cat ../.licenses_header.md > $[[ inputs.remote_license_file ]]
```
danielskinstad, line 100: Do you think it would be worth it adding some 'mkdir -p' to the dir in case it doesn't exist (the case for mender-gateway)

danielskinstad (approved):
LGTM. Some questions/comments, but feel free to merge once Lluis' comment is adressed

templates/release-oslicenses-golang.yml
```diff
@@ -0,0 +1,106 @@
...
+      when: on_success
+  image: golang:$[[ inputs.golang_version ]]
+  needs: []
+  variables:
+    GOFLAGS: $[[ inputs.golang_flags ]]
+  before_script:
+    - go install github.com/google/go-licenses@v1.6.0
+    - if [ ! -f go-licenses.gotpl ]; then
+    -   curl --fail --output go-licenses.gotpl
+          https://raw.githubusercontent.com/mendersoftware/mendertesting/refs/heads/add-golang-license-template/go-licenses.gotpl
```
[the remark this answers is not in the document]
elkoniu: Good catch - fixed

elkoniu:
@mender-test-bot start client pipeline

elkoniu:
@mender-test-bot sync

templates/release-oslicenses-golang.yml
lluiscampos: @danielskinstad I think it is a good idea, actually. We didn't have it for the changelog (there we will fail if the dir or file does not exist) and that was our reference when implementing this component. But certainly it will ease the work for adding the release to new repositories (and we will always review it as humans, so that is fine).

@elkoniu But let's not do this right now here. We can follow-up on both components (changelog and license) afterwards on a separate PR.

PR #398, opened 2025-09-17
QA-1166: Automatically tag and publish the release

lluiscampos:
Based on feedback from the demo, we can save one more click by allowing this job to happen automatically.


oldgiova (approved):

elkoniu (approved):
LGTM but I did not mind having this "human factor" here :)

PR #399, opened 2025-09-17
ci: Adjust and align PR titles and bodies for two components

lluiscampos:
These two components do something very similar, but they were using different messaging. Additionally the changelog one was mentioning `CHANGELOG.md` which does not exist.


lluiscampos:
For reference, these are the created PRs with the old code:
* https://github.com/mendersoftware/mender-docs-changelog/pull/144
* https://github.com/mendersoftware/mender-docs/pull/2653

PR #461, opened 2026-04-09
fix: Update sleep period in changelog creation

elkoniu:
release:candidate:update-changelog shall run after release:candidate:tag-and-release job is completed.


danielskinstad (approved):
The change LGTM, but I would just modify the commit message a bit. 
```
release:candidate:update-changelog shall run after release:candidate:tag-and-release
job is completed 
```
turned out to not be true.


Both the PR and git-cliff stuff happens in the same job, so the issue is the race condition _within_ the job

PR #465, opened 2026-04-13
ci: Bump golang image for test:static

elkoniu:
Test static is useing and external dependency which requires image to be at least `golang:1.22`

### DIFF
--- a/.gitlab-ci-check-golang-static.yml
+++ b/.gitlab-ci-check-golang-static.yml
@@ -27,7 +27,7 @@ test:static:
   needs: []
   except:
     - /^saas-[a-zA-Z0-9.]+$/
-  image: golang:1.20
+  image: golang:1.22
   before_script:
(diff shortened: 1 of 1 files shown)


.gitlab-ci-check-golang-static.yml
```diff
@@ -27,7 +27,7 @@ test:static:
   needs: []
   except:
     - /^saas-[a-zA-Z0-9.]+$/
-  image: golang:1.20
+  image: golang:1.21
```
mzedel: > Test static is useing and external dependency which requires image to be at least golang:1.22

so should it not be:

```suggestion
  image: golang:1.22
```
?

mzedel (approved):

.gitlab-ci-check-golang-static.yml
elkoniu: Updated :)

PR #473, opened 2026-04-28
fix: fixing gitlab update status

p-targowicz:
Fixing gitlab update status

Ticket: QA-1505

### DIFF
--- a/.gitlab-ci-github-status-updates.yml
+++ b/.gitlab-ci-github-status-updates.yml
@@ -14,14 +14,18 @@ stages:
 .github_status_template:
   tags:
     - k8s-small
-  # Keep overhead low by using a small image with curl preinstalled.
-  image: ${CI_DEPENDENCY_PROXY_DIRECT_GROUP_IMAGE_PREFIX}/curlimages/curl-base
+  # Keep overhead low by using a small image with curl and git preinstalled.
+  image: registry.gitlab.com/northern.tech/mender/mender-test-containers:base-alpine-master
   retry:
     max: 2
     when:
       - runner_system_failure
       - stuck_or_timeout_failure
   before_script:
+    - |
+      _pr_sha=$(git rev-parse HEAD^2 2>/dev/null || true)
+      _sha="${_pr_sha:-${CI_EXTERNAL_PULL_REQUEST_SOURCE_BRANCH_SHA:-$CI_COMMIT_SHA}}"
+      GITHUB_STATUS_API_URL="https://api.github.com/repos/mendersoftware/$CI_PROJECT_NAME/statuses/$_sha"
     - |
       send_status() {
         local status="$1"


danielskinstad (approved):

mzedel (commented):
I don't think we have `CI_EXTERNAL_PULL_REQUEST_SOURCE_BRANCH_SHA` available since we handle the sync ourselves instead of the built in repo sync.

danielskinstad:
> I don't think we have CI_EXTERNAL_PULL_REQUEST_SOURCE_BRANCH_SHA available since we handle the sync ourselves instead of the built in repo sync.

That was my worry as well, but I tried `echo ${CI_EXTERNAL_PULL_REQUEST_SOURCE_BRANCH_SHA}` in https://github.com/mendersoftware/mendertesting/pull/472/changes/ff2e86ab11f0cb87f925e2e8531011759e14838f. See https://gitlab.com/Northern.tech/Mender/mendertesting/-/jobs/14129882620

```
$ echo ${CI_EXTERNAL_PULL_REQUEST_SOURCE_BRANCH_SHA}
ff2e86ab11f0cb87f925e2e8531011759e14838f
```

p-targowicz:
Yeah you guys might be right the only thing I'm worried is that if I do `git reset --hard HEAD^2` we would get back to what we started with - meaning that it'd run on `pull/NNN/head`  again but I think there's no other easy option to fix this thing

danielskinstad:
So it looks like you're partially right, @mzedel. It won't be set if we manually trigger a pipeline for a PR, only when the pipeline is created with mender-test-bot or through the github-gitlab sync

mzedel:
> only when the pipeline is created with mender-test-bot or through the github-gitlab sync

to me that sounds good, as that should cover the regular process IMO! and very nice to know that the var is set, thanks for checking @danielskinstad!

p-targowicz:
> So it looks like you're partially right, @mzedel. It won't be set if we manually trigger a pipeline for a PR, only when the pipeline is created with mender-test-bot or through the github-gitlab sync
> 
> > I'm worried is that if I do git reset --hard HEAD^2 we would get back to what we started with - meaning that it'd run on pull/NNN/head
> 
> Deleting the merge commit is not the same as running it on pull/N/head

Yeah not exactly however very close either way I can do add this, I'll leave it to you guys - I feel you know better than me 🥇

mzedel (approved):
let's see how we can work with this 🤞

danielskinstad:
> as that should cover the regular process IMO!

Almost, except for manual triggers which is quite common.

What if just get the sha from `git rev-parse  HEAD^2`? That should work for everything, right?

p-targowicz:
> > as that should cover the regular process IMO!
> 
> Almost, except for manual triggers which is quite common.
> 
> What if just get the sha from `git rev-parse HEAD^2`? That should work for everything, right?

im not sure i follow @danielskinstad what do you eman manual triggers - if means goind to gitlab and running pipeline "on demand" then would anything mater ? you pass there usually `your branch` and it should be fine no ? do i miss anything ?

danielskinstad:
What I mean is: you want to run a manual pipeline for this PR, so you go to https://gitlab.com/Northern.tech/Mender/mendertesting/-/pipelines, press `New pipeline` and select branch `pr_473`. Since the pipeline is not triggered by a command or the github-gitlab sync, `CI_EXTERNAL_PULL_REQUEST_SOURCE_BRANCH_SHA` will be empty, meaning we won't see the status of that pipeline in this PR.

p-targowicz:
oh ok running on `pr_x` got it, thank you ! we can do then `git rev-parse  HEAD^2` I guess

danielskinstad (approved):

PR #476, opened 2026-04-30
Renovate component

danielskinstad:
I am about to track a golang version in mender-dist-packages with renovate, so I figured it would be better to have a shared component rather than having to do a bunch of stuff to set it up each time


danielskinstad:
@mender-test-bot sync

oldgiova (approved):
Ah, so you can make a template out of a template? Very smart!

vpodzime (approved):
ACK.

PR #478, opened 2026-05-04
fix(coverage): unset GITLAB_CI before Codecov upload

p-targowicz:
codecov uploader auto-detects GitLab CI and forces \`service=gitlab\`, ignoring the \`--service github\` flag this causes uploads to be associated with GitLab MRs instead of GitHub PRs, so the Codecov app never posts coverage comments

Ticket: QA-1574


p-targowicz:
@oldgiova another PR for same thing sadly there's no way to test it w/o merging change to master .... thus I need to do it that way - sorry for inconvenience

oldgiova (approved):
Strange but... 🤷‍♂️

PR #479, opened 2026-05-04
fix(coverage): pass --sha explicitly to Codecov uploader

p-targowicz:
## Summary

Follow-up to #478.

After unsetting `GITLAB_CI` to prevent service auto-detection, the Codecov uploader can no longer infer the commit SHA from GitLab env vars, causing uploads to fail with `Unable to detect SHA and slug`.

Fix: pass `--sha ${CI_COMMIT_SHA}` explicitly so the uploader has everything it needs without relying on GitLab CI auto-detection.

Ticket: QA-1574


p-targowicz:
https://github.com/mendersoftware/mendertesting/pull/485

PR #484, opened 2026-05-12
Bump to latest go 1.26 in go templates

danielskinstad:
PRs like https://github.com/mendersoftware/mender-setup/pull/67 require newer golang versions (https://gitlab.com/Northern.tech/Mender/mender-setup/-/pipelines/2506546710)

We don't pin every golang repo to an old golang version, as only the components with recipes in meta-mender need to maintain backwards compatibility with the go version in Yocto. This means that some repos bump dependencies and therefore require newer go versions


vpodzime (approved):
ACK

michalkopczan:
ACK, thanks.

PR #488, opened 2026-06-17
ci: move all workloads to k8s

mzedel:


oldgiova (approved):
Thanks for taking care of this!

PR #489, opened 2026-06-18
rely on default retries, runner + k8s runners switch pt2

mzedel:


oldgiova (approved):
Thanks for taking care of this!

PR #493, opened 2026-06-24
chore: Update commit.config.js name

rewanrashid-boop:
Changed from .js to .cjs as a result of reference error regarding ES module scope

Ticket: QA-1561


danielskinstad:
Are we sure this won't break anything? mender-server (and -enterprise) has its own config:  https://github.com/search?q=org%3Amendersoftware%20commitlint.config.js&type=code
Maybe we can just check for both, so first check for .cjs and then .js

danielskinstad (approved):

PR #497, opened 2026-07-01
ci: Deprecate non multiplatform jobs in template

rewanrashid-boop:
Legacy jobs in template need to be deprecated. At the same time the repos still using these jobs need to be updated.

Ticket: QA-1487

### DIFF
--- a/.gitlab-ci-check-docker-build.yml
+++ b/.gitlab-ci-check-docker-build.yml
@@ -148,15 +148,8 @@ build:docker:
   before_script:
     - *export_docker_vars
   script:
-    - *requires-docker-common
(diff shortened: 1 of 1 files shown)


rewanrashid-boop:
This needs review and in the case of it being merged then the changes can propagate and these can test properly:
https://github.com/mendersoftware/mantra/pull/300
https://github.com/mendersoftware/mender-docs-site/pull/560

These are all updates to deprecate use of non multiplatform jobs following OPTION 2 in the task:
https://northerntech.atlassian.net/jira/for-you?tab=assigned&selectedIssue=QA-1487

oldgiova (approved):

PR #502, opened 2026-07-07
ci: make sync:image track-agnostic for multiplatform builds

oldgiova:
sync:image depended only on publish:image via `dependencies:`. Once a project disables publish:image in favor of publish:image-multiplatform (the pattern introduced by the previous two commits), that dependency job never runs, PUBLISH_IMAGE_DIGEST is never set, and sync:image silently commits a manifest with a blanked-out image: field to the target infra repo.

List both publish jobs under dependencies: so sync:image picks up whichever one actually ran, and add a guard that fails the job instead of pushing a broken manifest if the digest ends up empty anyway.

Ticket: QA-1487

Co-authored-with: Claude

### DIFF
--- a/.gitlab-ci-check-docker-deploy.yml
+++ b/.gitlab-ci-check-docker-deploy.yml
@@ -58,6 +58,7 @@ sync:image:
     - if: '$CI_COMMIT_BRANCH == "master"'
   dependencies:
     - publish:image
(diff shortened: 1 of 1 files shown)


rewanrashid-boop (approved):
Makes sense 🚀

frodeha (approved):

PR #504, opened 2026-07-08
feat(changelog): align git-cliff config with the new commit grammar

p-targowicz:
> [!WARNING]
> These configs are fetched from master at pipeline runtime, so this changes changelog output for **every release repo the instant it merges**. There is no gradual rollout — please give the team a heads-up before merging.

Rewrites both `utils/cliff.toml` and `utils/cliff.toml.scoped` (in one PR, so they cannot drift) to the new policy.

Per-commit rendering (QA-1670):
- Commit body is no longer rendered by default; opt in with `Changelog: Commit` (`Changelog: All` tolerated as a legacy synonym)
- `Changelog: <sentence>` replaces the rendered line; `Changelog: None` skips the commit (exact footer line only — a sentence starting with "None" still renders)
- Legacy `Changelog: Title` and miscased keyword values are silently ignored so historical ranges regenerate cleanly
- `Ticket: <id>` is appended to the line and auto-linked
- `Deprecation: <sentence>` populates a Deprecations section
- Dependency bumps split: CVE-flagged ones route to Security, the rest to a new Dependency updates section

Structure and ordering (QA-1671):
- Section order: Breaking changes → Deprecations → Security → New features → Improvements (perf + refactor) → Bug fixes → Dependency updates → "All tickets resolved" appendix (flat table); empty sections omitted
- Template hook (comment) for a manually curated Highlights block at the top
- `docs`/`style`/`test`/`ci`/non-bump `chore` commits no longer appear in the changelog (intentional, per the new section list)

Validation:
- Tested with git-cliff 2.13.1 against a 20-commit fixture covering every rule (trailer casing, subject collisions, CVE mentions in feat commits, deps-dev scopes, cherry-pick parentheticals, scopeless commits in the scoped config)
- Regenerated real changelogs from recent mender-server (direct config) and mender-artifact (scoped config) history; sections, ordering and routing verified

Merge order: after #503 (grammar spec), before the downstream doc repoint PRs.

Ticket: [QA-1670](https://northerntech.atlassian.net/browse/QA-1670)
Ticket: [QA-1671](https://northerntech.atlassian.net/browse/QA-1671)

[QA-1670]: https://northerntech.atlassian.net/browse/QA-1670?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #510, opened 2026-07-16
fix(changelog): end each release block with a trailing newline

p-targowicz:
Fast-follow fix for the config that went live in #504, reported on NorthernTechHQ/nt-gui#833: in a regenerated CHANGELOG, a release whose commits carry no tickets rendered its last bullet and the **next release's `##` header merged onto the same line** (`- Adjusted email endpoint## @northern.tech/store-0.41.0 - 2026-07-13`).

Root cause: git-cliff concatenates per-release body renders directly, and `trim = true` strips the template's own trailing newline — so a block only ended with a newline when the all-tickets appendix happened to render (its table rows carry one). Ticketless releases ended bare on the last bullet.

Fix: emit a trim-proof trailing newline (`{% raw %}`) at the end of the body template in both `cliff.toml` and `cliff.toml.scoped`, so every release block separates regardless of content.

Reproduced the nt-gui case locally (three-tag repo, middle release ticketless) — merged line before, clean separation after. Full 24-commit fixture suite passes for both configs.

Ticket: [QA-1670](https://northerntech.atlassian.net/browse/QA-1670)

[QA-1670]: https://northerntech.atlassian.net/browse/QA-1670?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/utils/cliff.toml
+++ b/utils/cliff.toml
@@ -97,6 +97,8 @@ body = """
 
 No Changelog found.
 {% endif %}
+{%- raw %}
+{% endraw -%}
 """
 # template for the changelog footer
 footer = """---"""
--- a/utils/cliff.toml.scoped
+++ b/utils/cliff.toml.scoped
@@ -110,6 +110,8 @@ body = """
 
 No Changelog found.
 {% endif %}
+{%- raw %}
+{% endraw -%}
 """
 # template for the changelog footer
 footer = """---"""


p-targowicz:
this is issue that @mineralsfree discovered

mineralsfree (approved):
Looks good to me 🙈

